PR #1, opened 2020-06-16
GOBMCN2-98 Build hostname from IP address if not specified as a parameter

muresan:


muresan:
The name.internal format is better than 172.16.1.1.internal where the dots confuse various tools and the hostname will end up being 172.

PR #22, opened 2020-06-23
Markdown updates to user guide

mfielding:
* Adding syntax highlighting for commands
* Ran through the linter and update with findings
* Update _config to use the GitHub markdown processor for GitHub pages
* Cleaning up <> quasi-HTML tags that is misinterpreted in the table


docs/user-guide.md
```diff
@@ -780,24 +794,24 @@ need a different value. The parameter values that the toolkit uses are echoed
 during execution so you can confirm the configuration.
 
 The complete list of parameters and their values are provided in the [Parameters
-section](#heading=h.m1e7sxtn4kzm).
+section](#parametersn).
```
bpj146, line 797: I think there's an extraneous "n" and the end of "#parameters"

bpj146: And by the way, great catch on these incorrect internal links. They are leftover from the Google doc. I should have caught them.

docs/user-guide.md
```diff
@@ -780,24 +794,24 @@ need a different value. The parameter values that the toolkit uses are echoed
...
 
 Configuring an installation is discussed in [Configuring
```
bpj146, line 799: I know you didn't change this, but I think this sentence and its link should be deleted, because it is kind of self referential.

docs/user-guide.md
mfielding: Sentence removed, thanks.

docs/user-guide.md
```diff
@@ -779,25 +793,22 @@ Most parameters have default values, so you only need to specify them when you
 need a different value. The parameter values that the toolkit uses are echoed
 during execution so you can confirm the configuration.
 
-The complete list of parameters and their values are provided in the [Parameters
-section](#heading=h.m1e7sxtn4kzm).
-
 Configuring an installation is discussed in [Configuring
 installations](#configuring-installations).
```
bpj146, line 797: I was suggesting that you delete the sentence "Configuring an installation is discussed in [Configuring 
installations](#configuring-installations)."

I think the sentence about the list of parameters is okay to leave as-is

mfielding: Done.

PR #24, opened 2020-06-24
Create a doc index page

mfielding:
This is a placeholder index page, redirecting to the user guide for now.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,4 +1,4 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Toolkit for Bare Metal Service/</title>
+<title>Toolkit for Bare Metal Services</title>
 <meta http-equiv="refresh" content="0; URL=https://google.github.io/bms-toolkit/user-guide.html">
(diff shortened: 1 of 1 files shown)


docs/index.html
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Toolkit for Bare Metal Service/</title>
```
bpj146, line 3: Is that trailing slash on "Service/" supposed to be there?

mfielding: Nope, fixed.

docs/index.html
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Toolkit for Bare Metal Services</title>
```
bpj146, line 3: Sorry I didn't notice this last time (I didn't get enough sleep last night): "Services" should be "Solution"

mfielding: Updated :-)

PR #30, opened 2020-07-13
GOBMCN2-129-2 - change hosts to use dbasm group

pythiangoryunov:
check-instance.yml
-- change hosts to use dbasm group

install-sw.yml
-- change hosts to use dbasm group
-- add condition DG for cluster_type

prep-host.yml
-- change hosts to use dbasm group

roles/base-provision/tasks/main.yml
-- add condition DG for cluster_type
-- correct IP address in /etc/hosts entry

roles/lsnr-create/tasks/main.yml
-- ignore errors since 
   it can not handle a change if firewall disabled
-- add tag

### DIFF
--- a/roles/lsnr-create/tasks/main.yml
+++ b/roles/lsnr-create/tasks/main.yml
@@ -19,7 +19,11 @@
     permanent: yes
     immediate: yes
     state: enabled
-  ignore_errors: true
+  become: yes
+  register: firewall_output
+  failed_when:
+    - "'firewall is not currently running' not in firewall_output.msg"
+    - "'Permanent and Non-Permanent(immediate) operation' not in firewall_output.msg"
   tags: lsnr-create
 
 - name: Test whether port is free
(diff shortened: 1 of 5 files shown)


roles/lsnr-create/tasks/main.yml
```diff
@@ -19,6 +19,8 @@
     permanent: yes
     immediate: yes
     state: enabled
+  ignore_errors: true
```
mfielding, line 22: ANy background on why we'd want to ignore errors?  Perhaps put as a comment in the file?

pythiangoryunov: if firewall is not running, the task fails saying it should be enabled. with offline: yes attribute (https://docs.ansible.com/ansible/latest/modules/firewalld_module.html) it was the same, so I did not find a way to ignore the task if firewall service is not running, so put ignore errors

mfielding: How about using failed_when (fail when we get a response code, and we don't see the string that might identify if firewalld isn't running)?

https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html

pythiangoryunov: ok, put several conditions to avoid cases when firewall is not enabled and when firewall is already configured for the port

PR #35, opened 2020-07-28
GOBMCN2-144 - DR - duplicate from backup

pythiangoryunov:
config-db.yml
-- rename tasks to main
-- change tags

roles/common/defaults/main.yml
-- add force_backup to force backup for standby duplication
   (default false)
-- add duplicate_type (active - default, backup) variable
   to choose type of standby duplication

roles/db-copy/tasks/main.yml
-- rename active-copy playbook
-- add functionality to take full backup on primary if it forced
-- copy backupsets to standby site using file system
-- add password copy after duplicate once again to avoid
   issues with DG creation

roles/db-copy/templates/duplicate.cmd.j2
-- add ability to duplicate using backup
-- add parameters for listener port

roles/db-copy/templates/initaux.ora.j2
-- add parameter for CDB

### DIFF
--- a/config-db.yml
+++ b/config-db.yml
@@ -47,13 +47,13 @@
 
   - include_role:
       name: db-copy
-      tasks_from: active-copy
+      tasks_from: main
     when:
       - create_db|bool
       - cluster_type != "RAC"
       - lookup('env', 'PRIMARY_IP_ADDR') is defined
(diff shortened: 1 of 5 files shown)


pythiangoryunov:
Not to merge for now, saved it for the future

PR #36, opened 2020-08-05
Add command-line argument for swap disk

mfielding:
The Oracle installer requires swap to exist, so it can be convenient to
create swap as part of installation.  Formerly this was done only on GCE
VMs, and had a hard-coded default of /dev/sdg.  Here instead, we remove
the default (which may not match the install location), and permit swap
creation on any platform.  This parameter is optional, and, if not
specified, swap will not be created at all.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -1079,6 +1079,17 @@ NTP_PREF
 Optional: set only if you need to manually define an NTP server, instead of
 relying on the OS defaults.</td>
 </tr>
+<tr>
+<td>Swap device</td>
+<td><p><pre>
+SWAP_BLK_DEVICE
+--swap-blk-device
+</pre></p></td>
+<td>user defined - no default</td>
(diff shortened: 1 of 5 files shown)


mfielding:
> I would make swap definition as part of json with file systems definition rather than another option of command line, however possibly it is a matter of preferences.

To your comment (and apologies if we've talked about this before):  JSON files, as currently implemented, are just shell scripts calling jq to extract parameters.  Since they don't use Ansible's variable handling at all, they can't be easily used with Ansible tower and other higher-level automations.  So as a user, I do like the ability to pass parameters four ways:  via group_vars file, ansible's extra vars, install-oracle.sh environment variable, or install-oracle.sh parameter.  Granted, we could probably refactor to avoid needing to list parameters in 4 different places (install-oracle.sh's parameters, install-oracle.sh validation, , group_vars, and user guide)

PR #38, opened 2020-08-30
GOBMCN2-151 - backup destination

pythiangoryunov:
install-oracle.sh
-- restrict backup destination to start either with "/" or "+"

roles/common/defaults/main.yml
-- correct paths for home directories of users

roles/ora-host/tasks/main.yml
-- create backup directory on FS
-- check ASM DG in config file if backup points to ASM


pythiangoryunov:
I updated "Cloud Filestore" to "file system destination"

mfielding:
> I updated "Cloud Filestore" to "file system destination"

But is this check even necessary now?  By requiring a leading + or /, there's now no way to get an empty backup destination.

pythiangoryunov:
--backup-dest is mandatory parameter and checking + or / is to avoid something else to be put in as relative path ("BACKUP" for example). And might be --backup-dest should not be a mandatory since installation script can skip database config

mfielding:
> --backup-dest is mandatory parameter and checking + or / is to avoid something else to be put in as relative path ("BACKUP" for example). And might be --backup-dest should not be a mandatory since installation script can skip database config

So what message should a user expect when they don't specify a backup destination?

pythiangoryunov:
as of now backup destination is mandatory so it is not possible to skip it

mfielding:
After this change, what message will a user see if they don't specify a
backup destination?

pythiangoryunov:
possibly nothing should be shown, but if no backup destination then no backup taken and the question if the scripts should be copied and configured on a host and if database should be changed to archivelog

mfielding:
> possibly nothing should be shown, but if no backup destination then no backup taken and the question if the scripts should be copied and configured on a host and if database should be changed to archivelog

Wouldn't line 626 show an error message?  (and thus line 721 never run)?

pythiangoryunov:
yes, it will be "Incorrect parameter provided for backup-dest: " message from line 626

mfielding:
Then line 729 is dead code, right?

>

pythiangoryunov:
yes, you are right! the changes I made avoid message notification "Please specify a Cloud Filestore or disk group location".
Do you want to change to make the parameter optional? and if it is specified it should be a message that it should be either full path for file system or disk group?

mfielding:
My original comment was only to make sure we clean up dead code, but now
that you mention it, making the backups optional could help some use cases
that cannot store backups due to limited disk space or something else.  So
if it isnt't too much work, that might be worthdoing, but don't
hesistate to use a different change for that.

>

pythiangoryunov:
I will work on it and will submit as different pull request

pythiangoryunov:
I removed lines with dead code

mfielding (approved):

PR #44, opened 2020-09-22
GOBMCN2-162 - backup-dest is optional

pythiangoryunov:
install-oracle.sh
-- make backup-dest optional
-- ask for backup-dest if no skip-database-config option is used

### DIFF
--- a/install-oracle.sh
+++ b/install-oracle.sh
@@ -230,8 +230,8 @@ COMPATIBLE_RDBMS="${COMPATIBLE_RDBMS:-0}"
 COMPATIBLE_RDBMS_PARAM="^[0-9][0-9]\.[0-9].*"
 
 ###
-GETOPT_MANDATORY="ora-swlib-bucket:,backup-dest:"
-GETOPT_OPTIONAL="ora-version:,no-patch,ora-edition:,cluster-type:,cluster-config:,ora-staging:,ora-db-name:,ora-db-domain:,ora-db-charset:,ora-disk-mgmt:,ora-role-separation:"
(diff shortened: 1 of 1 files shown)


pythiangoryunov:
The change to make backup-dest optional. However if --skip-database-config option is not used (then play config-db is in place) then backup-dest is asked

mfielding (commented):
Although we change the enforcement here, do we need a tag or some other way to actually avoid trying to create the backup dest and backup creation?  Or will that be skipped implicitly if we don't use the config-db playbook?

pythiangoryunov:
further on if backup destination starts with "/" it will be created in ora-host role
if it starts with "+" (meaning it should point to disk group) it is checked that the disk group exists

from roles/ora-host/tasks/main.yml
...
- name: Create backup directory
  file:
    path: "{{ backup_dest }}"
    state: directory
    owner: "{{ oracle_user }}"
    group: "{{ oracle_group }}"
    mode: "ug=rwx"
  when: backup_dest[0:1]  == "/"
  tags: os-dirs
- name: Check ASM backup destination
  set_fact:
    dg_exists: "{{ asm_disks | json_query('[?diskgroup==`' + backup_dest[1:] + '`].diskgroup') }}"
  when: backup_dest[0:1]  == "+"
  tags: backup-dest

- name: Fail if ASM backup diskgroup not defined
  fail:
    msg: Diskgroup assigned for backup should be defined for creation
  when:
    - backup_dest[0:1] == "+"
    - dg_exists|length == 0
  tags: backup-dest

mfielding (approved):

PR #47, opened 2020-09-22
T-165 Make instance-hostname mandatory to avoid using IP address as hostname

muresan:
Make instance-hostname mandatory to avoid using IP address as hostname


mfielding (commented):
Any background here?  Why do we need a hostname?

This change would cause a lot of existing scripts and documentation to fail, not the last of which is the example in the README.

muresan:
This PR was based on the request:

"If instance IP or hostname isn't specified, can we error out rather than having a hard-coded 192.168.56.201 default?" 

We've had a previous discussion about building some valid hostname but I don't remember a conclusion. There is no 192.168.56.201 default, maybe that's a Vagrantfile default somewhere, but not in the repo.

I agree that this PR is incomplete, it needs doc/README/examples update but we need to agree on how to proceed further.

muresan:
@googlebot I signed it!

mfielding:
Given that there's (no longer?) a default IP, I think we can reasonably close this out, as that was my original concern.  Thanks!

PR #48, opened 2020-09-22
T-178 Added role to open listener/scan ports for RAC

muresan:


mfielding (commented):
What about the case where the host has no firewall?  Given that the config should be functional in this case, I'm not sure if we want to raise a fatal error halfway through in such a case.  Maybe just a warning?

muresan:
The intial request was to "open a firewall port for the listener in RAC as for single instance". We already open a listener port for the single instance setup, this adds a similar set of tasks to have the same ports opened but based on scan_port not on listener port.
By default, based on disable_firewall variable, we do not disable the firewall. In order for this task to not fail, I'll make it dependent on that variable so we do not try to open ports in the firewall if it's already disabled.

muresan:
@googlebot I signed it!

muresan:
> What about the case where the host has no firewall? Given that the config should be functional in this case, I'm not sure if we want to raise a fatal error halfway through in such a case. Maybe just a warning?

I added conditionals to ensure task is executed only when firewall is enabled. Default is firewall enabled.

mfielding (approved):

PR #49, opened 2020-09-22
Added -g to ntpd to allow clock set regardless of time difference

muresan:
Ensured "-g" option is present to ensure NTP syncs regardless of time difference.
From the man page: 

Normally, ntpd exits with a message to the system log if the offset exceeds the panic threshold, which is 1000 s by default. This option allows the time to be set  to  any  value without  restriction;  however,  this can happen only once.

reordered tasks to start NTP after configuration files are created

### DIFF
--- a/roles/base-provision/tasks/main.yml
+++ b/roles/base-provision/tasks/main.yml
@@ -49,11 +49,11 @@
     lock_timeout: 180
   tags: os_packages
 
-- name: Ensure "-g" is present in /etc/sysconfig/ntpd
+- name: Update OPTIONS in /etc/sysconfig/ntpd
   lineinfile:
     dest: /etc/sysconfig/ntpd
     regexp: '^OPTIONS='
-    line: 'OPTIONS="-g"'
+    line: 'OPTIONS="{{ ntp_options }}"'
   notify: restart ntpd
 
 - name: Add ntp preferred server
--- a/roles/base-provision/tasks/main.yml
+++ b/roles/base-provision/tasks/main.yml
@@ -49,12 +49,12 @@
     lock_timeout: 180
   tags: os_packages
 
-- name: Make sure ntp is running
-  service:
-    name: ntpd
-    state: started
-    enabled: yes
-  tags: ntp
+- name: Ensure "-g" is present in /etc/sysconfig/ntpd
+  lineinfile:
+    dest: /etc/sysconfig/ntpd
+    regexp: '^OPTIONS='
+    line: 'OPTIONS="-g"'
+  notify: restart ntpd
 
 - name: Add ntp preferred server
   blockinfile:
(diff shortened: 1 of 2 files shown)


roles/base-provision/tasks/main.yml
```diff
@@ -49,12 +49,12 @@
...
-  service:
-    name: ntpd
-    state: started
-    enabled: yes
-  tags: ntp
+- name: Ensure "-g" is present in /etc/sysconfig/ntpd
+  lineinfile:
+    dest: /etc/sysconfig/ntpd
+    regexp: '^OPTIONS='
+    line: 'OPTIONS="-g"'
```
mfielding, line 56: I think this line would clobber existing options; when I check a test RAC host though, -g is the only option set anyway.  If there's an easy way to preserve existing options though that would be ideal.

mfielding (approved):

roles/base-provision/tasks/main.yml
muresan: "-g" is the option to have here, I see it's the default, I cannot reproduce what you saw, NTP not setting the correct time on the 1st run, regardless of time difference, it worked for me on 20+ years difference.

```
[root@localhost ~]# date --set="1 Jan 2000"
Sat Jan  1 00:00:00 UTC 2000
[root@localhost ~]# date
Sat Jan  1 00:00:17 UTC 2000
[root@localhost ~]# service ntpd start
Redirecting to /bin/systemctl start ntpd.service
[root@localhost ~]# ntpq -np
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
 162.159.200.1   10.21.8.19       3 u    -   64    1   32.064  6541657   0.645
 78.141.31.73    85.199.214.98    2 u    1   64    1   68.572  6541657   0.000
 185.103.117.60  194.80.204.184   2 u    -   64    1   38.097  6541657   0.000
 85.199.214.99   .INIT.          16 u    -   64    0    0.000    0.000   0.000
[root@localhost ~]# ntpq -np
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
 162.159.200.1   .STEP.          16 u   27   64    0    0.000    0.000   0.000
 78.141.31.73    .STEP.          16 u  949   64    0    0.000    0.000   0.000
 185.103.117.60  .STEP.          16 u  242   64    0    0.000    0.000   0.000
 85.199.214.99   .STEP.          16 u    -   64    0    0.000    0.000   0.000
[root@localhost ~]# date
Wed Sep 23 08:42:52 UTC 2020
[root@localhost ~]# ps xfa|grep ntp
 3577 pts/0    S+     0:00                          \_ grep --color=auto ntp
 3570 ?        Ss     0:00 /usr/sbin/ntpd -u ntp:ntp -g
[root@localhost ~]#

```

I'll update to not clobber OPTIONS

muresan: There's no easy solution to not clobber, other than OPTIONS from read file, modify it and write it back (3 extra tasks), all other options are not simple/elegant.
In the end I do not see NTP configuration clobber as an issue, we already are very opinionated about configuring NTP with specific NTP servers so if someone wants to ensure certain options are available in /etc/sysconfig/ntpd then they can be set explicitly.

I've updated the PR to expose ntp_options as a parameter so we are authoritative on what OPTIONS from /etc/sysconfig/ntpd is set to

mfielding: Looks good, thanks!

mfielding (approved):

mfielding:
Catalin, can you please sign the CLA?

I've fixed the branch protection to enforce this properly in the future.

PR #52, opened 2020-10-20
GOBMCN2-171 - additional parameters for cleanup script

pythiangoryunov:
cleanup-oracle.sh
-- add parameters ora-disk-mgmt,ora-swlib-path,ora-staging,
   ora-asm-disks,ora-data-mounts


cleanup-oracle.sh
```diff
@@ -30,6 +30,23 @@ ORA_VERSION_PARAM='^(19\.3\.0\.0\.0|18\.0\.0\.0\.0|12\.2\.0\.1\.0|12\.1\.0\.2\.0
...
+
+ORA_SWLIB_PATH="${ORA_SWLIB_PATH:-/u01/swlib}"
+ORA_SWLIB_PATH_PARAM="^/.*"
+
+ORA_STAGING="${ORA_STAGING:-""}"
+ORA_STAGING_PARAM="^/.+$"
+
+ORA_ASM_DISKS="${ORA_ASM_DISKS:-asm_disk_config.json}"
+ORA_ASM_DISKS_PARAM="^.*\.json$"
+ORA_ASM_DISKS_PARAM="^.*$"
```
mfielding, line 44: Doesn't this override the previous line?  I'd think we'd want to list ORA_ASM_DISKS_PARAM once.

cleanup-oracle.sh
```diff
@@ -30,6 +30,23 @@ ORA_VERSION_PARAM='^(19\.3\.0\.0\.0|18\.0\.0\.0\.0|12\.2\.0\.1\.0|12\.1\.0\.2\.0
...
+
+ORA_DATA_MOUNTS="${ORA_DATA_MOUNTS:-data_mounts_config.json}"
+ORA_DATA_MOUNTS_PARAM="^.*\.json$"
+ORA_DATA_MOUNTS_PARAM="^.*$"
```
mfielding, line 48: Similarly here:  ORA_DATA_MOUNTS_PARAM defined and then changed.

cleanup-oracle.sh
```diff
@@ -55,13 +72,33 @@ while true; do
         shift;
         ;;
     --inventory-file)
-	INVENTORY_FILE="$2"
+      	INVENTORY_FILE="$2"
```
mfielding, line 75: Spaing looks a bit off here, and on the next line too.

cleanup-oracle.sh
```diff
@@ -76,6 +113,10 @@ while true; do
     shift
 done
 
+[[ "$ORA_STAGING" == "" ]] && {
+    ORA_STAGING=$ORA_SWLIB_PATH
```
mfielding, line 117: Please quote; though I'd hope nobody would put a space in the pathname here :-)

pythiangoryunov: swlib path quoted before, I assume it should be okay

pythiangoryunov: --ora-swlib-path)
        ORA_SWLIB_PATH="$2"
        shift;
        ;;

cleanup-oracle.sh
pythiangoryunov: removed and added spaces back

cleanup-oracle.sh
pythiangoryunov: removed those with json - having extension does not apply any strict checks on format so it does not matter what extension name of the file

cleanup-oracle.sh
pythiangoryunov: removed those with json - having extension does not apply any strict checks on format so it does not matter what extension name of the file

install-oracle.sh
```diff
@@ -158,15 +158,12 @@ ORA_RECO_DISKGROUP="${ORA_RECO_DISKGROUP:-RECO}"
 ORA_RECO_DISKGROUP_PARAM="^[a-zA-Z0-9]+$"
 
 ORA_ASM_DISKS="${ORA_ASM_DISKS:-asm_disk_config.json}"
-ORA_ASM_DISKS_PARAM="^.*\.json$"
```
mfielding, line 161: I'm not sure we want to lose the parameter validation entirely here.  If the intent is to permit either a blank response or a ".json", something like (^$|\.json$) might work

pythiangoryunov: I changed it to "^.+\.json$" - if no value provided default json file name taken otherwise it should be file with json extension

mfielding (approved):

PR #53, opened 2020-10-20
GOBMCN2-25 added controlnode-provision to provision a control node instance

muresan:


muresan:
Closed the original PR: https://github.com/google/bms-toolkit/pull/12

mfielding (commented):
I'm not sure this is something we want to run by default;  users may very well choose to have a NAT gateway rather than a squid proxy, for example, or to configure DNS resolution via cloud DNS inbound forwarding rather than dnsmasq.  Is there a way for this code to be available but somewhat standalone?  I've seen a contrib/ directory on some projects, but there may be a cleaner way too.

muresan:
> I'm not sure this is something we want to run by default; users may very well choose to have a NAT gateway rather than a squid proxy, for example, or to configure DNS resolution via cloud DNS inbound forwarding rather than dnsmasq. Is there a way for this code to be available but somewhat standalone? I've seen a contrib/ directory on some projects, but there may be a cleaner way too.

The code is there as a role but it's not included anywhere. I can move it to contrib if it feels more appropriate.

mfielding:
Sure, I think a separate directory like contrib would be great, as it makes it easier to distinguish the roles that are indirectly called by install_oracle.sh from those that would need to be invoked explicitly.

PR #54, opened 2020-10-20
GOBMCN2-24 Added checks to ensure as much as possible that the main playbook wil…

muresan:
…l not fail

### DIFF
--- a/check-instance.yml
+++ b/check-instance.yml
@@ -21,8 +21,6 @@
         that: "ansible_version.full is version_compare('2.8', '>=')"
         fail_msg: "You must update Ansible to at least 2.8 to use these playbooks"
         success_msg: "Ansible version is {{ ansible_version.full }}, continuing"
-  vars:
-    allow_install_on_vm: false
 
   roles:
     - check-instance
--- a/check-instance.yml
+++ b/check-instance.yml
@@ -24,49 +24,5 @@
   vars:
     allow_install_on_vm: false
 
-  tasks:
-
-    - name: Test connectivity to target instance via ping
-      ping:
-      register: pingrc
-
-    - name: Abort if ping module fails
-      assert:
-        that: "pingrc.ping == 'pong'"
-        fail_msg: >-
-          The instance does not have an usable python distribution
-        success_msg: >-
-          The instance has an usable python installation, continuing
-
-    - name: Collect facts from target
-      setup:
- 
-    - name: Platform verification
-      fail:
-        msg: "The system may not be provisioned on VM by default. Use the allow_install_on_vm option for VM testing"
-      when:
-        - ansible_system_vendor == "Google"
-        - ansible_virtualization_type == "kvm"
-        - ansible_virtualization_role == "guest"
-        - allow_install_on_vm == false
-
-    - name: Test privilege escalation on target
-      raw: sudo -u root whoami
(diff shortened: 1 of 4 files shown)


muresan:
The PR has the same changes as https://github.com/google/bms-toolkit/pull/6

roles/check-instance/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+allow_install_on_vm: false
+
+ansible_user: ansible
+
+required_cpu_cores: 4
```
mfielding, line 19: Why 4?  I've successfully run oracle on a single core without too much trouble

roles/check-instance/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
...
+required_cpu_arch: x86_64
+
+required_mem_mb: 15000
+
+supported_distributions_19c:
```
mfielding, line 24: I'm concerned this list gets very complicated quickly, and challenging to keep up to date, as more versions and OSes are added.  For example, SuSE Linux, el8 kernels, Oracle 12cR1, 11gR2, 20c etc?  Might this be one place we could fall back on Oracle's own install prerequisite checks?

check-instance.yml
```diff
@@ -24,49 +24,5 @@
   vars:
     allow_install_on_vm: false
```
mfielding, line 25: This was added to roles/check-instance/defaults/main.yml;  is it possible to remove here?

mfielding (commented):
Thanks, I like the refactoring here.

roles/check-instance/defaults/main.yml
muresan: This is a pre-install check, for obvious reasons we cannot include here the Oracle checks and the initial request was to eliminate as much as possible an install-time fail, this is the OS check part. The OS/Kernel combination is compared to the list and anything newer is accepted so the list needs to be updated only once, anything newer will be automatically included.
I could not find anything for the other distributions so we can have a baseline and start from there, but I wanted something official

check-instance.yml
muresan: Will move.

roles/check-instance/defaults/main.yml
muresan: This came from the Oracle DBA team. I can remove or anyone can change to 1 but for BM there are no cases where this will fail.

muresan: I changed it to 1 so any combination is valid. Still the check is there, someone might change it to something else and use it.

roles/check-instance/defaults/main.yml
mfielding: Thanks for your work here.  My preference would be to avoid duplicating too much of the Oracle checks, but rather check for what we need to get the toolkit itself to run.  How about removing the kernel versions entirely, and just focusing on OS major releases, ideally without making distinctions between Oracle versions.  Perhaps as simpe as OL7 or RHEL7?  (And I hope OL8/RHEL8 too?)

PR #56, opened 2020-11-01
GOBMCN2-170 - different file systems for mounting across servers

pythiangoryunov:
docs/user-guide.md
-- description of ability to use ORA_DATA_MOUNTS_<hostname>
   to mount different devices

roles/host-storage/tasks/main.yml
-- add functionality to handle different files for mount points
   with usage of ORA_DATA_MOUNTS_<hostname> variables


pythiangoryunov:
functionality to mount different devices across servers using ORA_DATA_MOUNTS_<hostname> variables. I updated documentation as well

mfielding (commented):
I'm not a huge fan of using environment variables as a primary configuration mechanism;  they are persistent across runs, so can be mastakenly set and remain set for another run by mistake.  They also complicate invocation by cmobining configuration across command-line options, json file, and now also env variables.  Is there a way we can possibly avoid needing to use env variables here?  Like something in the jaon or even dbasm.yml itself

pythiangoryunov:
My understanding was that such particular configuration requirement is rare and only needed in some certain cases but not for massive installations that's why I chose the usage of the variables to point to specific files if needed and if variables not defined installation just takes the one defined as usual. 
However if it is not acceptable I would try and explore configuration within json files.

mfielding:
I'm going to close this really old PR;  the underlying issue still exists, but, on BMS< can be worked around using host provisioning.

PR #58, opened 2020-11-16
Update asm-create.yml

amadel8:
This role creates additional diskgroups but it doesn't mount them on other nodes, so diskgroups other than the first one created (usually data) are not mounted on nodes other than the master node. The proposed task is identical to "asm-create | (asmlib) Create disk groups" but it runs the srvctl start diskgroup command

### DIFF
--- a/roles/gi-setup/tasks/asm-create.yml
+++ b/roles/gi-setup/tasks/asm-create.yml
@@ -114,3 +114,21 @@
     msg: "{{ create_dg }}"
     verbosity: 1
   tags: gi-setup,start-asm
+  
+  - name: rac-asm-create | Mount Diskgroups
+  shell: |
+    srvctl start diskgroup -diskgroup {{ item.diskgroup }}
+  environment:
+    ORACLE_HOME: "{{ grid_home }}"
+    PATH: "{{ grid_home }}/bin:${PATH}"
+    ORACLE_VERSION: "{{ oracle_ver }}"
+    ORACLE_SID: "{{ asm_sid }}"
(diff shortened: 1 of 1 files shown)


mfielding:
Thanks for the detailed description!  This particular role doesn't run at all for RAC installs;  in RAC installs (hndled by roles/rac-gi-setup/*.yml).  I just ran a RAC installs on the existing code without errors;  looking at the installer and ASM logs, it happened something like this:

- Physical disks and /dev/asmdisks/ udev symlinks created
- GI installer runs, and creates the first diskgroup (where the voting disks sit).  Node 2 sees it on initial install.
- roles/rac-gi-setup/tasks/rac-asm-create.yml runs on node 1, and creates the second diskgroup, and mounts locally
- when DBCA registers the newly-created database with the OCR, it lists a hard dependency
<START_DEPENDENCIES:hard(global:uniform:ora.RECO.dg, global:uniform:ora.DATA.dg)
- The grid oraagent then mounts the diskgroup to satisfy the dependency
SQL> ALTER DISKGROUP RECO MOUNT  /* asm agent *//* {1:281:2417} */

This all being said, I don't see a specific harm from explicitly mounting the diskgroup;  we'd just want it to be in the RAC config.

amadel8:
Correct, I had an error after grid installation and before RAC role started and that's why I noticed the diskgroup is not mounted. This pull request can be ignored

amadel8:
Correct, I had an error after grid installation and before RAC role started and that's why I noticed the diskgroup is not mounted. This pull request can be ignored

PR #61, opened 2020-11-16
Update data_mounts_config.json

amadel8:
Add the attributes num, start, end and host_name will allow multiple options:
1- Creating /u01 on all RAC nodes by providing the device name on each host
2- Different devices on each node could be listed to create multiple mount points (/u01 and /u02 for example)
3- One device can be used to create multiple OS partitions and create different mount points, according to the attributes num, start and end it could be specified the size of each partition and its associated mount point

The new attributes will be used by the role host-storage (updated in another pull request)

### DIFF
--- a/data_mounts_config.json
+++ b/data_mounts_config.json
@@ -1,18 +1,26 @@
 [
     {
         "purpose": "software",
-	"blk_device": "/dev/BOGUS",
+	"blk_device": "/dev/mapper/<WWID>",
 	"name": "u01",
+	"num": "1",
+        "start": "0%",
+        "end": "100%",
 	"fstype":"xfs",
 	"mount_point":"/u01",
-	"mount_opts":"defaults"
+	"mount_opts":"defaults",
+	"host_name":"svr001"
     },
     {
         "purpose": "diag",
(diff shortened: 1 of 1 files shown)


data_mounts_config.json
```diff
@@ -1,18 +1,26 @@
 [
     {
         "purpose": "software",
-	"blk_device": "/dev/BOGUS",
+	"blk_device": "/dev/mapper/<WWID>",
 	"name": "u01",
+	"num": "1",
```
mfielding, line 6: I think these partition changes probably belong to PR #62 where they are discussed

data_mounts_config.json
```diff
@@ -1,18 +1,26 @@
...
-	"mount_opts":"defaults"
+	"mount_opts":"defaults",
+	"host_name":"svr001"
     },
     {
         "purpose": "diag",
-	"blk_device": "/dev/BOGUS",
-	"name": "u02",
+	"blk_device": "/dev/mapper/<WWID>",
+	"name": "u01",
```
mfielding, line 17: Two different mounts both called u01?

PR #62, opened 2020-11-16
Update tasks in role host-storage

amadel8:
To utilize a modified data_mount_config.json a few changes to make here:

- I don't think we should use LVM here, also searching for mapper in the device string doesn't make sense because of multipath disks. 
- We will create the partition only on the host that matches the hostname in data_mount_config.json, we can create multiple partitions on the same device
- OEL adds the partition number to the device name while RHEL adds 'p' followed by partition number hence checking for the distribution to add filesystem and mount properly

### DIFF
--- a/roles/host-storage/tasks/main.yml
+++ b/roles/host-storage/tasks/main.yml
@@ -16,10 +16,12 @@
 - name: Partition Oracle user mount devices
   parted:
     device: "{{ item.blk_device }}"
-    number: 1
+    number: "{{ item.num }}"
+    part_start: "{{ item.start }}"
+    part_end: "{{ item.end }}"
     state: present
     label: gpt
-  when: "'mapper' not in item.blk_device"
+  when: "ansible_hostname==item.host_name"
   with_items:
     - "{{ oracle_user_data_mounts }}"
   tags: ora-mounts
@@ -27,17 +29,8 @@
 - name: Add filesystem on Oracle user mount raw devices
   filesystem:
     fstype: "{{ item.fstype }}"
-    dev: "{{ item.blk_device }}1"
-  when: "'mapper' not in item.blk_device"
-  with_items:
-    - "{{ oracle_user_data_mounts }}"
-  tags: ora-mounts
-
-- name: Add filesystem on Oracle user mount LV devices
-  filesystem:
-    fstype: "{{ item.fstype }}"
-    dev: "{{ item.blk_device }}"
-  when: "'mapper' in item.blk_device"
+    dev: "{{ item.blk_device }}{{ '' if ansible_distribution == 'OracleLinux' else 'p' }}{{ item.num }}"
+  when: "ansible_hostname==item.host_name"
   with_items:
     - "{{ oracle_user_data_mounts }}"
   tags: ora-mounts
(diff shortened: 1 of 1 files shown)


mfielding:
One of the benefits of SN-based storage like BMS (or cloud PD) provides is that individual LUNs can be created and resized as needed for each mount.  On the other hand, partition tables have several issues:
(a) They can't easily be resized, especially when there is more than one partition
(b) The OS doesn't consistently update partition tables when underlying disks are mounted/removed/remounted
(c) Partition tables can hurt performance due to cylinder alignment issues
(d) As you've discovered, different operating systems treat partition tables differently;  you've discovered one difference, but there may be more across other OS versions and perhaps even storage vendors
LVM can be high-maintenance as well but doesn't suffer from the issues above (except for maybe b)

roles/host-storage/tasks/main.yml
```diff
@@ -16,52 +16,33 @@
...
-  with_items:
-    - "{{ oracle_user_data_mounts }}"
-  tags: ora-mounts
-
-- name: Add filesystem on Oracle user mount LV devices
-  filesystem:
-    fstype: "{{ item.fstype }}"
-    dev: "{{ item.blk_device }}"
-  when: "'mapper' in item.blk_device"
+    dev: "{{ item.blk_device }}{{ '' if ansible_distribution == 'OracleLinux' else 'p' }}{{ item.num }}"
```
jcnars, line 32: Greetings !

Trying to understand this PR in detail.
For `{{ item.blk_device }}` can you please provide some examples of such device names and configs (additional details) where it doesn't contain `mapper` and the corresponding storage layout in the BMS host?

Maybe output from commands such as lsblk, fdisk / parted, multipath, sanlun lun show, pvs, vgs, lvs could paint a picture of the layout. (Of course assuming, you will be scrambling out the WWIDs and other such details to sanitize!)

Also, the data_mounts_config.json that was fed into `install-oracle.sh` will be helpful as well.

(I see changes to that json file proposed in the other PR #61, but it will be good to have the file that worked with the changes to host-storage/tasks/main.yml file)

Thanks

amadel8: Hello JC,

The main point of this change is allowing the toolkit to create and configure /u01 and /u02 as local mount points on RAC when every machine has its own local LUN and hence we need a way to assign LUN WWIDs to server names. Please check section "4- Edit Ansible Configuration json Files" in the following document to clarify all possible scenarios that could be achieved by modifying data_mounts_config.json
https://docs.google.com/document/d/1MRwg0kUWhTxn62GQ0-xD7HRCqVgDZJiABp4qIc_kki8/edit?usp=sharing

jcnars: Hello Ahmad,
Thanks for the doc and all your inputs on this PR #62 and PR #61.

We have identified and tested the solution for a variety of similar scenarios for local host-storage including udev-ising and kpartx-ing the devices so that any local host device presented is utilized correctly for mountpoints like, say `/u01`.

While are working on the design, coding, testing of the scenarios with you in the loop, you may pls close the loop on this PR #62 and PR #61.

Thanks

PR #67, opened 2021-04-05
Allow brute-cleanup to work with RAC

mfielding:
brute-cleanup was originally designed for single-node, Oracle Restart installs;  on RAC clusters I ran into a few issues and making a few fixes:

- Making existing service shutdown and CRS deconfigure run only on non-RAC clusters (detected by the presence or absence of a cluster_name in the inventory file)
- Adding a somewhat hacky RAC hard shutdown using the OHASD shutdown handler, plus rootcrs.sh.  Why not use Oracle's deinstall?  I'd prefer not to depend on a response file taht may or may not reflect current reality, but rather do as generic a deconfigure/shutdown as possible
- Adding tfa (trace file anaalyzer) to the kill list, as well as removing the initscript itself
- I ran into a case where umount failed on /u01 but lsof showed nothing in use.  a lazy (umount -l) seems to resolve by simply detaching the mount, and seems to fit the context of a "brute cleanup", especially since we're going to zero out the device later anyway.  Ansible's mount handler doens't support the option, so we're reverting to shell here, with the unfortunate side effect that nonexistant mounts will generated ignored error messages.


jcnars:
Tested the `cleanup` branch's `roles/brute-ora-cleanup/tasks/main.yml` file on a non-RAC host and confirm that the newly added commits successfully execute the intended functionality.

The run is documented in this gpaste (internal): 
https://paste.googleplex.com/6116459345346560

This LGTM from a non-rac standpoint.
Aiming to test this on a RAC cluster in the coming days.

jcnars:
Able to test run this multiple times successfully for a single-node RAC cleanup.
https://paste.googleplex.com/6236106396794880 (internal)

jcnars (approved):

PR #68, opened 2021-04-05
Default db_domain to an empty string

mfielding:
By default, dbca uses an empty string for db_domain.  The purpose of db_domain is to
disambiguate database names when there might be conflicts.  But the existing
default of "world" is no better than an empty string in this regard.  So to
avoid creating new defaults a user might not expect, here we change the default
db_domain to an empty string.

Sample output (internal): https://paste.googleplex.com/4736418472525824


jcnars (approved):
Reviewed referenced gpaste, lgtm

PR #70, opened 2021-04-08
Ignore errors adjusting sqlpatch objects

mfielding:
On a never-patched system, {{ oracle_base }}/cfgtoollogs/sqlpatch/sqlpatch_history.txt doesn't exist.  Rather than failing in this case, ignore the error and continue.


jcnars (approved):

jcnars (approved):
Reviewed referenced gpaste, lgtm

PR #74, opened 2021-05-17
Enforce hugepages with use_large_pages=ONLY

mfielding:
Currently the toolkit configures hugepages for database shared memory,
and sets use_large_pages=TRUE to enable their use.  However, future
post-install configuration, such as expansion of shared memory or
affecting memlock limits, can cause the database to allocate shared
memory from smaller pages.  This can be dangerous, not only because the
performance and page table size benefits of hugepages aren't achieved,
but also that they can cause both hugepages and non-hugepage memory to
be allocated, potentially resulting in out-of-memory situations.

By setting use_large_pages=ONLY, we tell the database to refuse to start
if it cannot allocate large pages as required.  This parameter is
supported all the way back to Oracle 11.2.0.2 [1], so should even work
with the 11.2.0.4 configuration.

[1]
https://docs.oracle.com/cd/E11882_01/server.112/e40402/initparams268.htm


jcnars:
LGTM

jcnars (approved):

PR #77, opened 2021-06-22
Zero out entire user data disk in cleanup

mfielding:
The existing zero-out logic appends "1" to the device name;  but if the device has no partitions, it results in a nonexistant device and the disk is not zeroed out at all.  Modifying the logic to unconditionally zero out the the first 100mb of the raw device, which covers both the partition table and any filesystem headers.  Then run partprobe to allow the kernel to clean up former partition devices, if any.


jcnars (approved):
LGTM

PR #78, opened 2021-06-25
Update Jinja and style on backup scripts

mfielding:
Update the RMAN backup and arch scripts in line with https://google.github.io/styleguide/shellguide.html as well as the shellcheck linter.  And, in order to simplify testing, move all the Jinja code to the top where substitutions are easier.

Key updates:
- Variables in lower case
- Consistent left-side formatting
- Removing $? checks: https://github.com/koalaman/shellcheck/wiki/SC2181
- Adding quoting when referencing variables
- Clarifying that a_date is really the start timestamp
- Jinja code limited to constant definitions
- Removing db_name references where the instance_name can be used instead

### DIFF
--- a/roles/db-backups/templates/rman_arch_backup.sh.j2
+++ b/roles/db-backups/templates/rman_arch_backup.sh.j2
@@ -11,6 +11,9 @@ ts="date +%Y-%m-%d_%H-%M-%S"             # Timestamp format
 start_ts="$($ts)"                        # Start timestamp
 log_dir="{{ logs_dir }}"                 # A directory for the logs
 type="ARCH"                              # Type is archivelog backup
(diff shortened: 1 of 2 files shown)


jcnars:
Wondering if these changes have been tested.

I usually try to test the codepaths that's changed by calling the playbook changed with the tag or if the usecase demanded, by doing a `install-oracle.sh` or `apply-patch.sh` , to ensure nothing breaks that may slip through a review.

mfielding:
jcnars@:  after some delay I do have a working cloud install (separate PR to come), where I could run the changes from the Ansible site.

Output:  https://gist.github.com/mfielding/0c49bec8b6b0948233b8913f83554ab2

jcnars (approved):
LGTM based on the test results

PR #82, opened 2021-07-20
Simplify and speed up gcscopy

mfielding:
The gcscopy.yml tasks use the gsutil cat command on the control node,
and stream it to the BMS host's software library.  This mulit-stage
process avoids a dependency on Internet access or gcloud authentication
on the BMS host iteslf, and instead uses the pre-configuraed state of a
cloud VM on the control node.  Additionally, gcscopy caches patch fiels
on the BMS host to avoid expensive re-downloads when not necessary.

To protect against gsutil cp failures and partially-written files,
"unzip -t" is used before and after to validate the archive.  This is a
slow and expensive process requiring reading multiple gigabytes of data.

In thie change, we replace the "unzip -t" with an "unzip -l", listing
the archive contents.  As zip directories are at the end of the file, it
provides good detection of partially-copied files with much reduced
overhead.

At the same time, we simplify the code by removing the ls step (which is
covered by the unzip), add changed_when logic that provides user
feedback when a re-download was required, and avoid the post-download
check, as we already get good coverage by ansible's implicit check or
return codes combined with the set -o pipefail inthe shell, which
captures errors in either the gsutil or cat commands; we do this by
looiing for the lack of unzip -l's headers, which would have indicated
an existing file was validated.

Unnecessary [[ $? -eq 0 ]] is removed from shell output, as per
https://github.com/koalaman/shellcheck/wiki/SC2181 recommendations.

Sample output:
https://gist.github.com/mfielding/9e6e33d051764e314017006e1fd20557

### DIFF
--- a/roles/swlib/tasks/gcscopy.yml
+++ b/roles/swlib/tasks/gcscopy.yml
@@ -17,35 +17,12 @@
   shell: |
     set -o pipefail
     {% for i in item.files %}
-    ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }} \
-      "ls {{ swlib_path }}/{{ i }}"
-    if [[ $? -eq 0 ]]
+    if ! ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }} \
+      "unzip -l {{ swlib_path }}/{{ i }}"
       then
-        ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }} \
-          "unzip -t {{ swlib_path }}/{{ i }}"
-        if [[ $? -ne 0 ]]
-          then
-            gsutil cat gs://{{ swlib_mount_src }}/{{ i }} | ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} \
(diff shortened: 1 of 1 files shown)


roles/swlib/tasks/gcscopy.yml
```diff
@@ -17,177 +17,96 @@
...
 
 # overwrite opatch zipfile on BMS host from gcs every time that may be cached currently in {{ swlib_path }}
 # reason to overwrite is that the opatch zipfile in MOS is named the same, example: p6880880_190000_Linux-x86-64.zip
 # this change downloads OPatch from GCS bucket every time with neglible overhead as opatch zip file is typically ~120M
 - name: gcscopy | Copy GI OPatch update files from GCS to target instance
   shell: |
     set -o pipefail
-    gsutil cat gs://{{ swlib_mount_src }}/{{ item.patchfile }} | ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} \
+    if ! ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }} \
+      "unzip -l {{ swlib_path }}/{{ item.patchfile }}"
```
jcnars, line 62: IIUC, the logic proposed here is:
if the opatch file (say: `p6880880_190000_Linux-x86-64.zip`) exists on the server and it is intact (as borne by the rc of the `unzip -l`), then we DO NOT copy the file from GCS.

I think we need to copy the opatch files every time from GCS, whether the BM host has it or not in the swlib path on the host.

So, the `if` condition may not be needed.

mfielding: Good catch;  I thought I had different logic here, but I guess I didn't.

https://gist.github.com/mfielding/07337b39ec43a3660aea8ed47b849882

jcnars (approved):
LGTM, thanks for making the code efficient and elegant !

PR #83, opened 2021-07-22
Simplify logic for non-multipath udev rules

mfielding:
The toolkit writes udev rules for non-device-mapper cases using blk_devices from the ASM disk JSON file, looking up the SCSI ID using the "ansible_device_links" facts. There are various rules for GCE, VirtualBox, and BMS that attempt to determine the orrect list element by static position [0] or [1].  Although this logic seems to be deterministic, it's complex and can be fragile across versions.  This change uses the Jinja2 select filter to match the `scsi-` list element instead of list position, and them removes the `scsi-` to get the SCSI ID itself, allowing us to use one task for all non-multipath cases.

Sample output: https://gist.github.com/mfielding/0b41604bb03c2574fdfa32aa285a25e6

### DIFF
--- a/roles/ora-host/tasks/main.yml
+++ b/roles/ora-host/tasks/main.yml
@@ -39,7 +39,7 @@
   tags: selinux
 
 - name: Ensure NOZEROCONF=yes is in /etc/sysconfig/network
-  lineinfile: 
+  lineinfile:
     dest: /etc/sysconfig/network
     regexp: '^NOZEROCONF'
     line: 'NOZEROCONF=yes'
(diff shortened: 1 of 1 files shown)


jcnars:
From the hosts I hve tested, the foll. old jinja expr (due to mpathed LUNs):
```bash
block: "KERNEL==\"sd?1\", SUBSYSTEM==\"block\", PROGRAM==\"/usr/lib/udev/scsi_id -g -u -d /dev/$parent\",
RESULT==\"{{ hostvars[inventory_hostname]['ansible_device_links']['ids']['%s' | format( item.1.blk_device | 
regex_replace('/dev/') )][0] | regex_replace('scsi-') }}\", 
SYMLINK+=\"{{ path_udev }}/{{ item.1.name }}\", 
OWNER=\"{{ grid_user }}\", GROUP=\"{{ grid_group }}\", MODE=\"0660\""

```
is usually skipped and the asm disks have the following rules (derived from an code that's untouched in this PR):
```bash
# BEGIN ASM disk udev rules for /dev/mapper/3600a098038314344382b4f7539397476
ACTION=="add|change", ENV{DM_UUID}=="mpath-3600a098038314344382b4f7539397476", 
SYMLINK+="asmdisks/DATA_7539397476", GROUP="asmadmin", OWNER="grid", MODE="0660"
```

The new expr with this PR:
```bash
block: "KERNEL==\"sd?1\", SUBSYSTEM==\"block\", PROGRAM==\"/usr/lib/udev/scsi_id -g -u -d /dev/$parent\", 
RESULT==\"{{ hostvars[inventory_hostname]['ansible_device_links']['ids']['%s' | format( item.1.blk_device | 
regex_replace('/dev/') )] | select('match','scsi-') | first | regex_replace('scsi-') }}\", 
SYMLINK+=\"{{ path_udev }}/{{ item.1.name }}\", OWNER=\"{{ grid_user }}\", GROUP=\"{{ grid_group }}\", MODE=\"0660\""
```

results in (based on the gist file attached):
```bash
KERNEL=="sd?1", SUBSYSTEM=="block", PROGRAM=="/usr/lib/udev/scsi_id -g -u -d /dev/$parent", 
RESULT=="36000c29c47ed8f61d6443f2033e0d8fd", SYMLINK+="asmdisks/DATA1", OWNER="oracle", 
GROUP="asmadmin", MODE="0660"
```

The data structure for the `ansible_device_links` fact looks like:
```json
            "ansible_device_links": {
                "ids": {
                    "dm-0": [
                        "dm-name-3600a098038314344382b4f753939746a",
                        "dm-uuid-mpath-3600a098038314344382b4f753939746a",
                        "lvm-pv-uuid-r7lvLB-SjVA-77wU-rGq6-roii-EBnR-gZyX28"
                    ],
                    "sdaz": [
                        "scsi-3600a098038314352512b50634b7a2d51",
                        "wwn-0x600a098038314352512b50634b7a2d51"
                    ],
                    "sdaz1": [
                        "scsi-3600a098038314352512b50634b7a2d51-part1",
                        "wwn-0x600a098038314352512b50634b7a2d51-part1"
                    ],
                    "sdaz2": [
                        "scsi-3600a098038314352512b50634b7a2d51-part2",
                        "wwn-0x600a098038314352512b50634b7a2d51-part2"
                    ],
                    "sdaz3": [
                        "scsi-3600a098038314352512b50634b7a2d51-part3",
                        "wwn-0x600a098038314352512b50634b7a2d51-part3"
                    ],
                },
```

And the scsi_id outputs as:
```bash
[root@linuxserver44 rules.d]# /usr/lib/udev/scsi_id -g -u -d /dev/sda
3600a098038314344382b4f7539397555

```
Based on the above, question for my understanding:
Are we doing this change, because it may not always be a given that scsi will be element [0] in the data structure for `ansible_device_links` excerpted above?
I see from the gist that this is tested. Any other place this needs to be tested (GCE/BMS)?

Also, it will help to see the error, if something failed specifically with current access via index [0].

mfielding:
Thanks for the detailed testing.  Indeed, the expression should return a SCSI ID based on the Ansible facts.  And the fact that we have we currently have three different rules with only a different array index talks to the complexity here, and I think it's definitely possible that orders may change in the future, as there are no guarantees in the spec.  The specific issue I had was that not every possible combination of hypervisor/guest was covered, so that we end up with no udev rule at all, which is tricky to debug.

I don't think we have any non-multipath BMS testcases, but I have tested it in GCE with success.  FWIW, the underlying facts look like this:

```
        "device_links": {
            "ids": {
                "sda": [
                    "google-persistent-disk-0",
                    "scsi-0Google_PersistentDisk_persistent-disk-0"
                ],
```

jcnars:
Thank you for the clarification and details.

jcnars (approved):
LGTM

PR #85, opened 2021-07-26
Default disk mount options to nofail

mfielding:
The toolit creates filesystem mounts specified in the
`--ora-data-mounts` parameter to `install-oracle.sh`, to store
contents like the ORACLE_HOME or diag dest.  A filesystem is
created here and added to /etc/fstab to mount on startup.  If
the filesystem is somehow not mountable, systemd prevents the
system from booting normally and enters emergency mode,
resulting in a hung system if no interactive console is
available.

This change adds a default "nofail" mount option, which
prevents the failure status from being propagated to systemd,
allowing the system to still boot normally, and the underlying
issue to be fixed.

Also fixing an issue with listener firewall to allow install to compelte.

Sample instal outputl: https://gist.github.com/mfielding/3867a17da4e01ff1cffea54f5c072a33

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Toolkit for installing and creating an initial Oracle database on Bare Metal Sol
             "name": "u01",
             "fstype":"xfs",
             "mount_point":"/u01",
-            "mount_opts":"defaults"
+            "mount_opts":"nofail"
         }
     ]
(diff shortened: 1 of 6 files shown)


jcnars (approved):
LGTM

PR #87, opened 2021-07-28
Refactor gcscopy and oracle_rel calculation

mfielding:
install-oracle.sh requires root access, not only on the database 
servers, but also the control node.  Not only is root acess on the 
control node a security issue, but it can use authentication issues with 
gsutil commands.  Removing this dpeendency requires a fairly complex 
refactoring to ensure that root privileges are available on the database 
hosts as required, but not when delegated to the control node 
(`delegate_to: 127.0.0.1`).  We now call swlib separately, only for RAC 
node 1 (if applicable), and without privilege elevation.

On a related note, there seem to be two different ways of calculating 
`oracle_rel`.  The calculation in grid infrastructure installs produces 
correct results, but the calculation in rdbms installs uses a lexical 
version comparison, so would consider version 18.2 to be higher than 
18.10, for example.  Unlike other variables, `oracle_rel` is complex 
enough that it cannot be set in `group_vars/all.yml`, and needs an 
actual playbook entry.  Here we create a playbook under common that can 
be called by all consumers:  GI/DB install, check-swlib, and patching.  
The logic has been updated to use pure Jinja2 and avoiding unnecessary 
computation and a dependency on external JSON libraries.

And lastly, a few "convenience" fixes encountered while developing and 
testing this:
- More work on cleanup, so that it (mostly) works even if the oracle or 
grid users are wrong.  Updating the process kills to use an ansibe loop, 
to avoid the process kill killing its own shell script wrapper.
- Doing two kill phases: SIGTERM and SIGKILL, much like the OS shutdown 
itelf.  This specifically addresses a case where Oracle TFA and 
OSWatcher can restart themselves during a kill process.
- Cerating DBCA resposne file as root, in case a previous run changed 
ora-role-separation settings.
- Found one yum call without a lock timeout, and fixed that, to work 
with concurrent cron jobs that occasionally run yum
- Better feedback on process kills, so ok/changed/fail are reported if 
no processes found, process killed, or unexpected error respectively
- Disabling `ANSIBLE_DISPLAY_SKIPPED_HOSTS` (except with when --debug is 
specified), considerably shortening the output and avoiding potentially 
confusing display of versions and code paths that are not appplicable to 
the current run.  
- Adding the same shorthand for versions to apply-patch.sh as 
install-oracle.sh has
- Defaulting swlib type to gcs, which is the standard example case

Design doc: go/bms-toolkit-gcscopy

Sample output:
- RAC: 
https://gist.github.com/mfielding/81cffc2df37ace6bfdde88d85434eccd
- Single instance: 
https://gist.github.com/mfielding/758a10c7508e6f2b7e7da0c4153b6851
- Patching: 
https://gist.github.com/mfielding/3779c61fd7b7c31f4b6d70b0fea28cb7 
- Cleanup: 
https://gist.github.com/mfielding/fbb7b90ac35a01611d822bcaa7a374c7
- Check-swlib: 
https://gist.github.com/mfielding/15bf29ddcf2043c2bcb023709555ddf6


apply-patch.sh
```diff
@@ -48,6 +48,7 @@ ORA_DB_NAME_PARAM="^[a-zA-Z0-9_$]+$"
 #
 INVENTORY_FILE="${INVENTORY_FILE:-./inventory_files/inventory}"
 
+export ANSIBLE_DISPLAY_SKIPPED_HOSTS=false
```
jcnars, line 51: Can we add comment something like (from the docs): 
```
# Override default Ansible behavior and avoid displaying skipped task/host entries in a task 
```

roles/brute-ora-cleanup/tasks/main.yml
```diff
@@ -61,25 +61,56 @@
...
-    pkill -9 ora_smon
-    pkill -9 tnslsnr
-    pkill -9 agent13c
-    pkill -9 -f PROTOCOL=beq
-    pkill -9 -f /grid/bin
-    pkill -9 -f tfa
-    pkill -9 -u {{ grid_user }}
-    pkill -9 -u {{ oracle_user }}
+- name: Kill (SIGTERM) any remaining Oracle processes
+  command: "/bin/pkill -f {{ item }}"
```
jcnars, line 65: Some distros have pkill under /usr/bin:

```
[jcnarasimhan@linuxserver44 ~]$ which pkill
/usr/bin/pkill
```

roles/db-create/tasks/rac-db-create.yml
```diff
@@ -91,7 +91,7 @@
 
 - name: rac-db-create | Create DBCA response file
   become: yes
-  become_user: "{{ oracle_user }}"
+  become_user: root
```
jcnars, line 94: Is this change to address any error?
I do not recalling facing any issues with creating rsp file as OS user `oracle`.

roles/patch/tasks/main.yml
```diff
@@ -39,14 +39,6 @@
...
   become_user: root
 
-- name: (GI) Copy patches from bucket
-  import_role:
-    name: swlib
-    tasks_from: gcscopy
-  vars:
-    patching_type: gi_patching
-  tags: patch-grid,sw-copy
-
```
jcnars, line 49: Thanks for testing this path, as I see in your gist !

apply-patch.sh
mfielding: Done.

roles/db-create/tasks/rac-db-create.yml
mfielding: The (albeit rare) issue is that `oracle_user` is a parameter and could potentially change between runs, such as changes to privilege separation.  Crating the file as root and chowning it allows any existing files to be clobbered.

roles/brute-ora-cleanup/tasks/main.yml
mfielding: Good catch.  Fixed.

jcnars (approved):
LGTM

PR #88, opened 2021-08-31
Host provision - bootstrapping actions done as `customeradmin` 

jcnars:
Ansible user creation and configuring ssh equivalence for the created user between the control node and DB hosts.

Successful validation: https://gist.github.com/jcnars/ff9d76574c9db4028a683faf0db6f8b1

Once this PR is reviewed and merged, will be using the same branch to submit other host provisioning actions 
(typically, those may no longer need the `customeradmin`)


roles/host-provision/tasks/user-setup.yml
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: user-setup | Create ansible user
+  user:
+    name: "{{ hostprovision_ansible_user }}"
+    uid: "2468"
+    comment: "Ansible toolkit host-provision automation"
```
mfielding, line 19: Please identify as bms-toolkit, even maybe a link https://github.com/bms-toolkit

roles/host-provision/tasks/user-setup.yml
```diff
@@ -0,0 +1,35 @@
...
+    groups: wheel
```
mfielding, line 20: hmmm;  is it safe to assume that this group will exist?  I also believe that, by default, the wheel group requires a (possibly nonexistant) local password to sudo.

An example of a way to maintain such a group;  to be even safer, we may want to consider to create a bms-toolkit specific group to avoid modifying the state of wheel on the control node. 
 https://stackoverflow.com/questions/33359404/ansible-best-practice-for-maintaining-list-of-sudoers

mfielding: No wait, I see you've built logic already to manage sudoers specifically for the user.  So I think the wheel group membership should be entirely unnecessary.

roles/host-provision/templates/hostprovision_ansible_user.j2
```diff
@@ -0,0 +1 @@
+{{ hostprovision_ansible_user }}      ALL=(ALL)      NOPASSWD: ALL
```
mfielding, line 1: Any chance we could get a quick comment in this file to note that it's for bms-toolkit runs, and is necessary for initial provisioning and patching only?

host-provision.sh
```diff
@@ -0,0 +1,109 @@
...
+       exit 1
+}
+
+# echo "PARSED COMMAND LINE FLAGS: $options"
+
+eval set -- "$options"
+
+while true; do
+    case "$1" in
+    --inventory-file)
```
mfielding, line 54: I'd really really like to avoid having to create yet another configuration file manually.  Can we please follow the existing logic in install-oracle.sh, taking parameters as required for host IPs (or a rac config file), and generating an inventory file from that?

host-provision.sh
```diff
@@ -0,0 +1,109 @@
...
+        INVENTORY_FILE="$2"
+        shift;
+        ;;
+    --hostprovision-ansible-user)
```
mfielding, line 58: From a user perspective, this naming is redundant;  we're asking for the user that bms-toolkit should run as.  Can we use the same parameter forms as isntall-oracle.sh has?

host-provision.yml
```diff
@@ -0,0 +1,64 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+# Assumption: 
+# First time connection is not through public private key pair
+# FIrst time connection is always through customeradmin credentials password authentication
```
mfielding, line 19: Can "custoemradmin" be parameterized?  No need for a command-line parameter, but something in group_vars/all.yml would help in case there's a case when the use is named something else.

roles/host-provision/tasks/ssh-keygen.yml
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Creates .ssh directory in control node if it does not exist
+  file:
+    path: ~/.ssh
```
mfielding, line 17: There are specific permission rquirements for the .ssh directory.  Can we set them here?

roles/host-provision/tasks/ssh-keygen.yml
```diff
@@ -0,0 +1,26 @@
...
+    state: directory
+  tags: host-provision 
+
+- name: ssh-keygen | Create pvt public key pair locally in control node
+  openssh_keypair:
+    # id_rsa is common name, name in a way to avoid overwriting it by chance
+    path: "~/.ssh/id_rsa_ansible_host_provision"
+    comment: "#generated by host-provision utility"
```
mfielding, line 25: Can we mention it's from bms-toolkit?

host-provision.sh
```diff
@@ -0,0 +1,109 @@
...
+out=`getopt -T`
+if [ $? != 4 ]; then
+  echo -e "Your getopt does not support long parameters, possibly you're on a Mac, if so please install gnu-getopt with brew"
+  echo -e "\thttps://brewformulas.org/Gnu-getopt"
+  exit
+fi
+
+# set -x
+GETOPT_MANDATORY="inventory-file:"
+# GETOPT_OPTIONAL="ora-role-separation:,ora-disk-mgmt:,ora-swlib-path:,ora-staging:,ora-asm-disks:,ora-data-mounts:,help"
```
mfielding, line 32: Remove commented lines?

jcnars: Done

roles/host-provision/tasks/ssh-keygen.yml
jcnars: Done

roles/host-provision/tasks/ssh-keygen.yml
jcnars: Done

host-provision.yml
jcnars: Done.

host-provision.sh
jcnars: Thanks for catching this!
Will change this to`--instance-ssh-user` keeping it consistent with from `install-oracle.sh`

host-provision.sh
jcnars: I see your point and agree with it.

My thoughts on this:
* For the `install-oracle.sh`, inventory variables and their parsing are important and we feed `--cluster-type RAC` and  `--cluster-config mntrl_cluster.json`.
    * the json file is dense with necessary info consumed by: `install-oracle.sh`.
* For the current `host-provision.sh`, we basically need the host IPs to connect to and there is no need to build an inventory (with the many host/group vars) at this stage of the host
   * consequentially, will explore the passing of IPs into `host-provision.sh` via cmd line and avoid having to create by hand another config file

Guess, this will satisfy the condition you pointed out of not having to hand-code and feed another config file...while, at the same time, keeping the overall design/flow simple.

roles/host-provision/templates/hostprovision_ansible_user.j2
jcnars: Done

roles/host-provision/tasks/user-setup.yml
jcnars: Thanks for the catch.
Removing `groups: wheel`

roles/host-provision/tasks/user-setup.yml
jcnars: Done

roles/host-provision/tasks/user-setup.yml
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: user-setup | Create ansible user
+  user:
+    name: "{{ hostprovision_ansible_user }}"
+    uid: "2468"
```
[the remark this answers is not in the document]
jcnars: Removing uid:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html#parameter-uid

roles/host-provision/tasks/ssh-keygen.yml
```diff
@@ -0,0 +1,26 @@
...
+- name: Creates .ssh directory in control node if it does not exist
+  file:
+    path: ~/.ssh
+    state: directory
+  tags: host-provision 
+
+- name: ssh-keygen | Create pvt public key pair locally in control node
+  openssh_keypair:
+    # id_rsa is common name, name in a way to avoid overwriting it by chance
+    path: "~/.ssh/id_rsa_ansible_host_provision"
```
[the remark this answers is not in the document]
jcnars: Naming it as `id_rsa_bms_toolkit` to indicate the algorithm used by openssh module and also parameterizing this file in group/all.yml as:
```
## host provision variables
hostprovision_ansible_user: "{{ lookup('env','INSTANCE_SSH_USER')|lower|default('ansible',true) }}"
hostprovision_firsttime_connect_user: "customeradmin"
hostprovision_key_file: "~/.ssh/id_rsa_bms_toolkit"
```

host-provision.yml
```diff
@@ -0,0 +1,64 @@
...
+        fail_msg: "You must update Ansible to at least 2.8 to use these playbooks"
+        success_msg: "Ansible version is {{ ansible_version.full }}, continuing"
+  tasks:
+  - include_role:
+      name: host-provision
+      tasks_from: user-setup.yml
+  remote_user: "customeradmin"
+  tags: host-provision
+
+- name: Validate ssh equivalence (passwordless connection) and sudo escalation for new user
```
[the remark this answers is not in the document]
jcnars: See your point and exploring this.
Thinking of moving this into the logic file as the very first task that does other things like LVM setup, etc.

host-provision.yml
```diff
@@ -0,0 +1,64 @@
...
+      tasks_from: ssh-keygen.yml
+  tags: host-provision
+
+- name: Create user and transfer public key to set up ssh equivalence
+  hosts: dbasm
+  vars_prompt:
+    - name: ansible_password
+      prompt: Enter customeradmin password
+  become: yes
+  pre_tasks:
```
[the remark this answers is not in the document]
jcnars: This is the earliest where a task (pre-task to be exact in this case) is run on the DB hosts.
(The prior task runs locally on the control node to generate the key files.)

group_vars/all.yml
```diff
@@ -131,3 +131,6 @@ install_rdbms: true
 disable_firewall: false
 
 ## End of non-modifiable variables section
+
+## host provision variables
+hostprovision_ansible_user: "{{ lookup('env','ORA_HOSTPROVISION_ANSIBLE_USER')|lower|default('ansible',true) }}"
```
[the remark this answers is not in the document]
jcnars: I think you are referring to the variable `INSTANCE_SSH_USER` that's populated in the main `install-oracle.sh` layer.

There this variable `INSTANCE_SSH_USER` was used at the bash level to build a inventory and after that not carried over into ansible level.

Hence re-using the `INSTANCE_SSH_USER` in the following assignment in `group/all.yml`

```yml
## host provision variables
hostprovision_ansible_user: "{{ lookup('env','INSTANCE_SSH_USER')|lower|default('ansible',true) }}"
hostprovision_firsttime_connect_user: "customeradmin"
hostprovision_key_file: "~/.ssh/id_rsa_bms_toolkit"
```

jcnars:
The inventory file creation has been avoided.

More structure introduced that will lend well into the next stage of introducing the core logic of host provisioning, like LVM, REHL subscription, etc.

host-provision.yml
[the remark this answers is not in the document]
jcnars: This was done, missed to mark it in the comment here.

host-provision.sh
jcnars: This has been done.

group_vars/all.yml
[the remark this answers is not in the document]
mfielding: I don't see these variables as being specific to host provisioning - they may be needed anywhere in the code.  A few suggestions:
instance_ssh_user
firsttime_connect_user
control_node_key_file

group_vars/all.yml
```diff
@@ -131,3 +131,8 @@ install_rdbms: true
 disable_firewall: false
 
 ## End of non-modifiable variables section
+
+## host provision variables
```
mfielding, line 135: I thought I put in a coment about this, but now can't find it.  Apologies if it's duplicated.

In terms of group vars, I'd prefer if we don't have hostprovision in the name, and instead have the ability to reuse them across multiple parts of the code, now or in the future.

A suggestion for naming:
instance_ssh_user
firsttime_connect_user
control_node_key_file

host-provision.sh
```diff
@@ -0,0 +1,104 @@
...
+
+eval set -- "$options"
+
+while true; do
+    case "$1" in
+    --instance-ssh-user)
+        INSTANCE_SSH_USER="$2"
+        shift;
+        ;;
+    --comma-separated-dbhosts)
```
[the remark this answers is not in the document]
jcnars: Thanks for the review and options.

Jotting my thoughts here:
* Per: https://github.com/google/bms-toolkit/blob/master/docs/user-guide.md#example-toolkit-execution, re-using `--instance-ip-addr` will be misleading and lead an end user more confused as to what the correct way is to specify based on what stage of the host (initial vs later installs patching) and what utility they are running and how they will need to specify it differently.
* The code triggered by `--instance-ip-addr` and `--comma-separated-dbhosts` cmd line switches is  different
* Dealing with inventory file creation at this stage (though a simple copy & paste from `install-oracle.sh`) will make the entry-point code bulkier and hard to maintain at 2 places.
* Calling the host-provision twice may be the most efficient thing (and least confusing), as runtime errors, if any, can be addressed on one host and then move on to the next one. Also helps with password prompt, you call out.
   * With this in mind, we can rename the variable `--comma-separated-dbhosts` to something like: `--dbhost_ip`

Please let me know your thoughts to proceed.

group_vars/all.yml
jcnars: Got it, will do

roles/host-provision/tasks/ssh-keygen.yml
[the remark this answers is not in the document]
jcnars: Ack on the variables renaming, will action on them, together with other changes that are needed after we are on the same page as to how to deal with the overall inventory file entity.

roles/host-provision/tasks/ssh-keygen.yml
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Creates .ssh directory in control node if it does not exist
+  file:
+    path: ~/.ssh
+    state: directory
+  tags: host-provision 
+
+- name: ssh-keygen | Create pvt public key pair locally in control node
```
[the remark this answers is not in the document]
jcnars: Ack, will do

group_vars/all.yml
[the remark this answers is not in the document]
jcnars: Ack, will do

host-provision.sh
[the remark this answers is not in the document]
jcnars: This proposal has been coded and tested successfully.
Pls check if it looks good.

Thanks

group_vars/all.yml
jcnars: All of the suggestions here have been incorporated and tested successfully.
https://gist.github.com/jcnars/d0e4887c736ddbd3dcc5c9a26d4d24db

Pls check if it looks good.

Thanks

host-provision.sh
```diff
@@ -27,7 +27,7 @@ if [ $? != 4 ]; then
   exit
 fi
 
-GETOPT_MANDATORY="comma-separated-dbhosts:"
+GETOPT_MANDATORY="dbhost_ip:"
```
mfielding, line 30: Sorry to be a pain here, but I don't think "dbhost" is a term we're using , but rather "instance-ip-addr" (and, separately, hyphens rather than underscores in arguments).  Can we be consistent with install-oracle.sh?

host-provision.yml
```diff
@@ -0,0 +1,71 @@
...
+  tags: host-provision
+
+- name: Create user and transfer public key to set up ssh equivalence
+  hosts: all
+  vars_prompt:
+    - name: ansible_password
+      prompt: Enter customeradmin password
+  vars:
+    #ansible_ssh_extra_args can be input as a command line or the following reasonable default may be used:    
+    ansible_ssh_extra_args: '-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentityAgent=no '
```
mfielding, line 44: On observation, `-o StrictHostKeyChecking=no` seems a bit dangerous from a security perspective.  https://docs.ansible.com/ansible/latest/user_guide/connection_details.html#managing-host-key-checking seems to indicate taht a user will be prompted interactively by default when an unknown hosts appears, which seems a bit more secure.  Would we be able to keep the defaults for host keys?

host-provision.yml
```diff
@@ -0,0 +1,71 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+# Assumption: 
+# First time connection is not through public private key pair
+# FIrst time connection is always through customeradmin credentials password authentication
```
mfielding, line 19: Nit: s/customeradmin/{{ firsttime_connect_user }}/

host-provision.yml
```diff
@@ -0,0 +1,71 @@
...
+  - include_role:
+      name: host-provision
+      tasks_from: ssh-keygen.yml
+  tags: host-provision
+
+- name: Create user and transfer public key to set up ssh equivalence
+  hosts: all
+  vars_prompt:
+    - name: ansible_password
+      prompt: Enter customeradmin password
```
mfielding, line 41: Nit: s/customeradmin/{{ firsttime_connect_user }}/

host-provision.sh
jcnars: Done

host-provision.yml
jcnars: Done

host-provision.yml
jcnars: done

host-provision.yml
jcnars: Removing this option

host-provision.sh
jcnars: Tested successfully:
```bash
✔ ~/DriveFS/My Drive/bmaas/host_provisioning/bms-toolkit [host-provision|✚ 2] 
15:48 $ runlocalssh ./host-provision.sh --instance-ip-addr 172.16.30.1  --instance-ssh-user ansible9
Command used:
/usr/local/google/home/jcnarasimhan/DriveFS/My Drive/bmaas/host_provisioning/bms-toolkit/host-provision.sh --instance-ip-addr 172.16.30.1 --instance-ssh-user ansible9

Running with parameters from command line or environment variables:

INSTANCE_SSH_USER=ansible9
INVENTORY_FILE=172.16.30.1,
ORA_CS_HOSTS=172.16.30.1

Found Ansible at /usr/bin/ansible-playbook

Running Ansible playbook: /usr/bin/ansible-playbook -i 172.16.30.1,   host-provision.yml

PLAY [Create private public key pair locally] **********************************************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************************************************************************
ok: [localhost]
...
```

host-provision.yml
jcnars: The latest round of changes have been incorporated and tested successfully; the following cmd line options is how it is now invoked:
```bash
✔ ~/DriveFS/My Drive/bmaas/host_provisioning/bms-toolkit [host-provision|✚ 2] 
15:48 $ runlocalssh ./host-provision.sh --instance-ip-addr 172.16.30.1  --instance-ssh-user ansible9
```

Please check from your end.
Thanks

host-provision.yml
```diff
@@ -38,10 +38,10 @@
   hosts: all
   vars_prompt:
     - name: ansible_password
-      prompt: Enter customeradmin password
+      prompt: Enter firsttime_connect_user password
```
mfielding, line 41: Is there a way to dereference the variable so that a user will see "Enter customeradmin password"

host-provision.yml
```diff
@@ -61,7 +61,7 @@
   vars:
     ansible_ssh_private_key_file: "{{ control_node_key_file }}"
     #ansible_ssh_extra_args can be input as a command line or the following reasonable default may be used:    
-    ansible_ssh_extra_args: '-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentityAgent=no '
+    ansible_ssh_extra_args: ' -o IdentityAgent=no '
```
mfielding, line 64: Nit: not sure the leading and trailing spaces are necessary

host-provision.yml
```diff
@@ -16,12 +16,12 @@
 
 # Assumption: 
 # First time connection is not through public private key pair
-# FIrst time connection is always through customeradmin credentials password authentication
+# FIrst time connection is always through {{ firsttime_connect_user }} credentials password authentication
```
mfielding, line 19: Nit: First, not FIrst

jcnars: Done

host-provision.yml
jcnars: It is true they are required, but I left the spaces as a placeholder/reminder, if in the future we get the `ansible_ssh_extra_args` from the command line and then stitch together a complex option (with many `-o`s), for example (this is a real-life example!):
```
    ansible_ssh_extra_args: '-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentityAgent=no -o ProxyCommand="ssh -W %h:%p -q jcnarasimhan@control-node.mfild.com"'
```

Will remove the spaces.

jcnars: Done

host-provision.yml
jcnars: There may be, but first glance, extrapolating this variable is introducing complexity that we can do away.

host-provision.yml
jcnars: Pushed a branch, but with this iteration, left the following as it is, as noted in an earlier comment:
 `      prompt: Enter firsttime_connect_user password`

We can revisit it later, if required.

jcnars: MOdified the prompt to:
```yml
      prompt: Enter customeradmin password
```

roles/host-provision/tasks/ssh-keygen.yml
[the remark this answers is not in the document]
jcnars: Done

host-provision.yml
mfielding: > MOdified the prompt to:
> 
> ```yaml
>       prompt: Enter customeradmin password
> ```

Adding background, it seems that `pause` runs too late in the execution step to be usable for SSH passwords, bringing us back to `vars_prompt`.  Change looks good.

mfielding (approved):

PR #95, opened 2021-11-18
creating the test_pr.sh that will be referenced by prowjob

jcnars:
Transferring shell commands used in https://gist.github.com/jcnars/26b8526cde45a1bb5778a36abe90b96b into this test script (context / reference: (internal) at: http://b/202240337#comment22)

This version can be considered still at POC stage and we will evolve more when we see results from prowjob.


presubmit_tests/test_pr.sh
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+### Transferring shell commands used in https://gist.github.com/jcnars/26b8526cde45a1bb5778a36abe90b96b
+### into this test script
+### context / reference: (internal) at: http://b/202240337#comment22
+
+# set up ssh from pod to BMX host
+# using sydney for initial testing
+mkdir /root/.ssh
```
mfielding, line 22: I would suggest:
- Isuing the install command to create/chmod in one command
- Using ~ in case this is ever run as non-root
- Removing argument from ssh-keyscan in case there is no ecdsa host key on the other host

bms_host=172.16.30.1

install -d -m 0700 ~/.ssh
ssh-keyscan "${bms_host}" > ~/.ssh/known_hosts

presubmit_tests/test_pr.sh
```diff
@@ -0,0 +1,35 @@
...
+chmod 0700 /root/.ssh
+ssh-keyscan -tecdsa 172.16.30.1 > /root/.ssh/known_hosts
+
+# install pre-reqs
+pip install jmespath
+cp /etc/files_needed_for_tk/google-cloud-sdk.repo /etc/yum.repos.d/google-cloud-sdk.repo
+yum install google-cloud-sdk -y
+
+# run the cleanup script
+cd ..; ./cleanup-oracle.sh --ora-version 19 --inventory-file /etc/files_needed_for_tk/nonrac-inv --yes-i-am-sure --ora-disk-mgmt udev --ora-swlib-path /u01/oracle_install --ora-asm-disks /etc/files_needed_for_tk/nonrac-asm.json --ora-data-mounts /etc/files_needed_for_tk/nonrac-datamounts.json
```
mfielding, line 32: I'd prefer an absolute path here

Given that you have most of the install and cleanup configuration, you might want to put the asm disk and other files here too, to avoid splitting test configuration over too many places

jcnars: I would have gone with absolute path if I am manually doing a `git clone` of the PR.
But since the prow is responsible for this clone, I don't know at this point at what location it will clone it. and hence I am not able to put the absolute path. We will change it to absolute once we do our first prowjob runs.

presubmit_tests/test_pr.sh
jcnars: Thanks, done

jcnars:
This is ready for your review, please.
Also, this works in tandem with another PR for prow, when you get a chance: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1306 so that this can be tested together.

presubmit_tests/test_pr.sh
mfielding: Where does stuff like nonrac-datamounts.json come from?  I think it might make sense to put the command-line arguments to cleanup/install-oracle.sh and the json files in the same place (this script, perhaps, or at least same directory/PR?).

And anit:  Can you use \ line continuation characters to keep line length reasonable?

jcnars: The `nonrac-datamounts.json` file comes from the GKE build cluster in `bmaas-testing` project's K8s configmaps.

https://github.com/GoogleCloudPlatform/oss-test-infra/blob/375bd5042646c04aff266a2bb82c397e18e5afeb/prow/prowjobs/google/bms-toolkit/test-install-oracle.yaml

has the details for files like `nonrac-datamounts.json` f, `nonrac-asm.json`, etc.
They all get mounted in the same place in the pod under: `/etc/files_needed_for_tk/`

(Internal) https://b.corp.google.com/issues/202240337#comment12 has the configmap creation steps.

2 points on the location of the `*.json` files:
1) They get mounted when the pod itself is created by our GKE build cluster
2) The cloning of the PR-ed repo happens after pod creation and hence it looks like at this point (before the first run), they cannot be combined in the same place.

As noted in previous comment, at this point, it's unclear where prow will clone the PR-ed repo. Once we have the first run, we can attempt to put everything in same place.

jcnars: Added line break \ to keep it readable

mfielding (approved):

PR #102, opened 2021-12-16
Add metadata for October 2021 patches

mfielding:
Adding patch metadata for Oracle's October 2021 PSUs, including 12.1,
12.2, and 19c.  11.2.0.4 and 18c no longer receive patchset updates.

The patch directory layout for GI/RDBMS combo patches (under
rdbms_patches) has changed, with an additional layer of nesting.  So
patch_siubdir now contains two directory levels.  To identify the
correct path, loko at bundle.xml and identify the singleton patch that
has all four components.  In case of multiple matches, one subdirectory
should be significantly larger; take that one.

PR #103, opened 2021-12-16
Fix -rolling argument to auatchauto apply

mfielding:
`opatch apply` takes a `-rolling` or `-nonrolling` argument, and the
patching code sets it to `-rolling` for RAC installs and `-nonrolling`
for non-RAC installs.  `opatchauto`, however, does _not_ take such an
argument and will fail with a cryptic 
`OPATCHAUTO-72050: Failed while retrieving system information.` 
error if the flag is given.

This issue appears to have bene introduced in the patch refactoring in
May 2021, and the fix is relatively simple:  don't add the argument for
`opatchauto` patches.

Note this PR is based on the octpsu branch from PR #102 

Sample output:
Before patch: https://paste.googleplex.com/6024404149469184?raw
After patch: https://paste.googleplex.com/5688927173214208?raw

### DIFF
--- a/roles/patch/tasks/rac-opatch.yml
+++ b/roles/patch/tasks/rac-opatch.yml
@@ -196,7 +196,7 @@
   tags: rac-opatch,stop-home
 
 - name: rac-opatch | Run GI opatch apply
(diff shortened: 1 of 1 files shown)


mfielding:
Discussing with jcnars@ live;  he points out that the default opatch (non-opatchauto) mode is `-rolling`, so we can apply a simpler patch:  `-nonrolling` for single instance, and defaults ('') otherwise.  This combination of arguments should work for both `opatchauto apply` and `opatch apply`

PR #107, opened 2022-06-05
Add metadata for January 2022 patches

mfielding:
Building on PR #102 we add patch metadata for January 2022 patches, based on https://www.oracle.com/security-alerts/cpujan2022.html and https://support.oracle.com/rs?type=doc&id=2817011.1


mfielding:
This change is part of the parent PR # 102

The commit in question:
https://github.com/google/bms-toolkit/pull/102/commits/150d1adc0ca399f7ba6bc95e72c4b664df15e40e

`opatch apply` takes a `-rolling` or `-nonrolling` argument, and the
patching code sets it to `-rolling` for RAC installs and `-nonrolling`
for non-RAC installs.  `opatchauto`, however, does _not_ take such an
argument and will fail with a cryptic `OPATCHAUTO-72050: Failed while
retrieving system information.` error if the flag is given.

This issue appears to have bene introduced in the patch refactoring in
May 2021, and the fix is relatively simple:  don't add the argument for
`opatchauto` patches.

Sample output:
Before patch: https://paste.googleplex.com/6024404149469184?raw
After patch: https://paste.googleplex.com/5688927173214208?raw

mfielding:
Closing in favor of #114

PR #111, opened 2022-06-16
Add troubleshooting for AnsibleUnsafeText errors

mfielding:
An error case users may run into:  errors in JSON file formatting.  The error message isn't very helpful, unfortunately.

### DIFF
--- a/docs/troubleshooting_guide.md
+++ b/docs/troubleshooting_guide.md
@@ -338,7 +338,7 @@ Thu Jan 21 22:40:31 UTC 2021
 Thu Jan 21 22:40:34 UTC 2021
 ```
 
-## Pass extra flags into ansible via `install-oracle.sh`:
+## Pass extra flags into ansible via `install-oracle.sh` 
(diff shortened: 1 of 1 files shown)


jcnars:
In the snippet you attached, is it safe to assume that it was a blank that caused this error?
>   fatal: [racnode1.orcl]: FAILED! => {"msg": "The conditional check ''mapper' not in item.blk_device' failed. The error was: error while evaluating conditional ('mapper' not in item.blk_device): 'ansible.utils.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'blk_device'\n\nThe error appears to be in '/home/mfielding/git/opatch/roles/host-storage/tasks/main.yml': line 16, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: Partition Oracle user mount devices\n  ^ here\n"}

jcnars (approved):

PR #112, opened 2022-06-16
Apply one off/OJVM patches separately from runInstaller

mfielding:
Oracle bug 32816171, introduced in 19.11, causes OJVM one-off patch
application from runInstaller's `-applyOneOffs` to error out with make
errors.  A year later, this bug has yet to be fixed.  So to permit us to
add new patch updates to the toolkit, this patch removes `-applyOneOffs`
from the `runInstaller` call, and instead runs `opatch` separately for each
applicable patch.

Although this bug appear to affect only OJVM patches in 19.11+, to avoid
complicating the code flow too much, this change modifies patch application
for all one-off patches.

More background on the issue:
https://mikedietrichde.com/2021/04/22/oracle-19c-installation-with-19-11-0-ru-ojvm-and-some-other-fixes/

### DIFF
--- a/roles/rac-db-setup/tasks/rac-db-install.yml
+++ b/roles/rac-db-setup/tasks/rac-db-install.yml
@@ -77,11 +77,6 @@
   become: yes
   tags: rac-db,sw-unzip
 
-- name: rac-db-install | Set variable for OneOff patch
-  set_fact:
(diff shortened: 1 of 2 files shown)


mfielding:
> If you can pls print out the patch install commands, that will be a record in the log file.

The command is actually logged in the output.  Example:

https://oss.gprow.dev/view/gs/bmaas-testing-prow-bucket/pr-logs/pull/google_bms-toolkit/112/bms-toolkit-install/1537542794883633152#1:build-log.txt%3A1355

            "/u01/app/oracle/product/19.3.0/dbhome_1/OPatch/opatch",
            "apply",
            "-silent",
            "-oh",
            "/u01/app/oracle/product/19.3.0/dbhome_1",
            "/u01/oracle_install/31720429/31668882"

jcnars (approved):

PR #113, opened 2022-06-22
Add metadata for October 2021 patches

mfielding:
Adding patch metadata for Oracle's October 2021 PSUs, including 12.1,
12.2, and 19c. 11.2.0.4 and 18c no longer receive patchset updates.

The patch directory layout for GI/RDBMS combo patches (under
rdbms_patches) has changed, with an additional layer of nesting. So
patch_siubdir now contains two directory levels. To identify the
correct path, loko at bundle.xml and identify the singleton patch that
has all four components. In case of multiple matches, one subdirectory
should be significantly larger; take that one.

(this is a replacement for PR #102)


jcnars:
/lgtm

This PR may not be needed? If the one with the latest PSU is merged, that should also have these changes?

jcnars (approved):
If merging, we may just merge the PR with APR 2022 patch. 

THis PR's changes , iirc, are also in that.

PR #122, opened 2022-08-23
Disable avahi only if the package exists, and fix 12cR2 RAC installs

mfielding:
As documented in Red Hat doc https://access.redhat.com/solutions/25463
and MOS note 1501093.1, the avahi daemon must be disabled on RAC
clusters.  But the existing disabling code fails if the avahi package
doesn't exist at all.

So here we add a "when" clause to only attempt to disable if the package
is not installed.  We do it via ansible's native builtin.packages
interface, and then using jinja2 syntax to check for a key in the
resulting dict.

A side effect of invoking builting.packages is taht it saves package
data to a variable called, plainly enough, `packages.  But the existing
toolkit code _also_ has a variable called packages, which gets ends up
getting clobbered.  To resolve this we rename `packages` to
`required_packages`.


jcnars:
I would also propose to add a `ignore_errors: yes` directive for the task: `- name: Disable avahi daemon`, as we don't want any failure of this for any other reason to be showstopper during install.

mfielding:
> I would also propose to add a `ignore_errors: yes` directive for the task: `- name: Disable avahi daemon`, as we don't want any failure of this for any other reason to be showstopper during install.

In this case, I'm trying to implement the narrowest possible fix for the problem we both encountered:  a case when the avahi package doesn't exist at all.  Telling ansbile to ignore all errors is a very "big hammer":  there could be real issues that need fixing.

jcnars (dismissed):

mfielding:
Merged the conflicting changes into this branch.

mfielding:
Created myself a RAC cluster to test this issue, and confirmed both issues still exist:

1. The cosmetic issue with ignored errors from avahi;  skipped in non-RAC installs.  So if it's the "happy path" and expected, it might be better to ignore.  Also renaming "packages" to "required_packages" to clarity.
2. The 12.2 installer failure when patching.  With `--no-patch` is works, but if patching up, it fails.

And found a third:

3. Failure of post-DB install opatch, because opatch is too old.  So we need to update opatch twice as part of the RDBMS install:  once before the RDBMS installer, and once after.  This case doesn't apply to all versions, but I figure it doesn't hurt to update even if not required.

Sample output:

* Installing 12cR2 without this change: install fails;  the specific error isn't in the log, but rather a generic failure  https://gist.github.com/mfielding/8014b9e693031a29b0bcb0ae0803ba4e

```
2025-02-20 21:56:44,849 p=70012 u=mfielding_google_com n=ansible | failed: [mfielding-ora122-node1 -> mfielding-ora122-node2(None)] (item=mfielding-ora122-node2) => {"ansible_loop_var": "item", "changed": true, "cmd": ["/u01/app/12.2.0/grid/root.sh"], "delta": "0:12:21.286072", "end": "2025-02-20 21:56:44.705983", "item": "mfielding-ora122-node2", "msg": "non-zero return code", "rc": 25, "start": "2025-02-20 21:44:23.419911", "stderr": "", "stderr_lines": [], "stdout": "Check /u01/app/12.2.0/grid/install/root_mfielding-ora122-node2_2025-02-20_21-44-23-427380457.log for the output of root script", "stdout_lines": ["Check /u01/app/12.2.0/grid/install/root_mfielding-ora122-node2_2025-02-20_21-44-23-427380457.log for the output of root script"]}
```

* Installing 12cR2 after the fixes for 1 and 2:  grid install succeeds, but post-install rdbms patching fails
https://gist.github.com/mfielding/93fe88914407ad5528ae6be1371c39c9
```
"Prerequisite check \"CheckMinimumOPatchVersion\" failed.", "The details are:", "", "", "The OPatch being used has version 12.2.0.1.6 while the following patch(es) require higher versions: ", "Patch 33561275 requires OPatch version 12.2.0.1.28.", "Please download latest OPatch from My Oracle Support.", "", "Log file location: /u01/app/oracle/product/12.2.0/dbhome_1/cfgtoollogs/opatch/opatch2025-02-21_23-25-16PM_1.log", "", "OPatch failed with error code 73"]}
```

* After adding the second opatch update, install succeeds
https://gist.github.com/mfielding/c0fc86981da66ba602e114cde5974bed

AlexBasinov (approved):

PR #125, opened 2022-08-25
Fix post-install DB opatch issues

mfielding:
When testing a 12.2 RAC install, which has an extra HAS patch, I ran into two issues with the opatch logic:

1. Unlike the GI install, the DB install is out of place, and runInstaller actually clobbers existing content in the `ORACLE_HOME`, including the OPatch executable.  Therefore we need to update OPatch _after_ the install has completed, but before invoking it to apply one-off patches
2. The opatch output display runs a loop over the appy_oneoff results, in order to handle multiple patches if necessary.  To display the output for the given patch, we need to print `{{ item.xxx }}` objects instead of the static `{{ apply_oneoff }}` ones.

### DIFF
--- a/roles/rac-db-setup/tasks/rac-db-install.yml
+++ b/roles/rac-db-setup/tasks/rac-db-install.yml
@@ -96,21 +96,6 @@
     dest: "{{ install_unzip_path }}/db_install.rsp"
   tags: rac-db,rac-db-response
 
-- name: rac-db-install | Update DB OPatch
-  unarchive:
-    src: "{{ swlib_path }}/{{ item.patchfile }}"
-    dest: "{{ oracle_home }}"
-    remote_src: yes
-  with_items:
-    - "{{ opatch_patches }}"
-  when:
-    - item.release == oracle_ver
-    - item.category == "OPatch"
-    - oracle_rel != "base"
-  become: yes
-  become_user: "{{ oracle_user }}"
-  tags: rac-db,update-opatch-db
-
 - name: rac-db-install | Set installer command
   set_fact:
     installer_command: "{{ runinstaller_loc }}/runInstaller -silent -waitforcompletion -responseFile {{ install_unzip_path }}/db_install.rsp {{ rel_patch | default('') }} {{ prereq_option }}"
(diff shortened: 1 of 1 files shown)


roles/rac-db-setup/tasks/rac-db-install.yml
```diff
@@ -153,8 +153,8 @@
 - name: rac-db-install | opatch output
   debug:
     msg:
-      - "{{ apply_oneoff.cmd }}"
-      - "{{ apply_oneoff.stdout_lines }}"
+      - "{{ item.cmd }}"
```
jcnars, line 156: The changes to this task `rac-db-install | opatch output` may be backed out, as a [recent PR](https://github.com/google/bms-toolkit/pull/126/files#diff-2d3864b3dd75f1a76901aadadafa95259779a756c66469f79c2c278812f6a481) for EL8.5 fixed it.

roles/rac-db-setup/tasks/rac-db-install.yml
```diff
@@ -137,6 +122,21 @@
       - "{{ install_rac_db }}"
   tags: rac-db,rac-db-install
 
+- name: rac-db-install | Update DB OPatch
```
jcnars, line 125: IIUC, the task block `rac-db-install | Update DB OPatch` has been moved in this PR after the task `rac-db-install | Set installer command`.

Upon eyeballing the pre & post blocks, it seems the content is same. Does the new ordering have any functionality effect?

roles/rac-db-setup/tasks/rac-db-install.yml
mfielding: Yup, I will merge into current HEAD once approved.

roles/rac-db-setup/tasks/rac-db-install.yml
mfielding: Indeed the order does matter.

Unlike the GI install, the DB install is out of place, and runInstaller actually clobbers existing content in the ORACLE_HOME, including the OPatch executable. Therefore we need to update OPatch _after_ the install has completed, but before invoking it to apply one-off patches

mfielding:
Incorporated a version of this fix into #122, as that install also failed due to the same reason.

PR #135, opened 2023-01-23
Remove diag mount from default data_mounts.json

mfielding:
Right now the toolkit doesn't actually use the `diag` mount for anything, so having it in the default config file is a bit confusing.  Removing it entirely.

### DIFF
--- a/data_mounts_config.json
+++ b/data_mounts_config.json
@@ -6,13 +6,5 @@
 	"fstype":"xfs",
 	"mount_point":"/u01",
 	"mount_opts":"nofail"
-    },
-    {
-        "purpose": "diag",
-	"blk_device": "/dev/BOGUS",
-	"name": "u02",
-	"fstype":"xfs",
-	"mount_point":"/u02",
-	"mount_opts":"nofail"
     }
 ]


jcnars:
/lgtm

While we are fixing this, can we also pls change the sample line from:
```
	"blk_device": "/dev/BOGUS",

```

to something like:
```
        "blk_device": "/dev/mapper/db-sw",  

```

maybe with an in-line comment such as:
\# `/dev/mapper/db-sw` is created if `host-provision.sh` utility was run...if that was not the case, use the correct block device that applies in your case

mfielding:
Thanks @jcnars for the suggestion.  Sadly, JSON doesn't allow comments at all.  And as you point out, we want this to work for people who haven't used host_provision.sh.  However the word BOGUS isn't great either.  What do you think about `replace_with_block_device_like_/dev/mapper/db-dw`  ?

jcnars:
> Thanks @jcnars for the suggestion. Sadly, JSON doesn't allow comments at all. And as you point out, we want this to work for people who haven't used host_provision.sh. However the word BOGUS isn't great either. What do you think about `replace_with_block_device_like_/dev/mapper/db-dw` ?

Thanks Marc, this does convey the example clearly, imo.

mfielding:
@jcnars PTAL, thanks.

jcnars (approved):

PR #137, opened 2023-01-24
Remove jinja2 references from when clauses

mfielding:
Current ansible versions give a warning:

```
TASK [rac-gi-setup : rac-asm-create | Check available ASM disk groups] *********
ok: [toolkit-ol8]
[WARNING]: conditional statements should not include jinja2 templating
delimiters such as {{ }} or {% %}. Found: created_dg is not search('{{
item.diskgroup }}')
```

The fix is simple:  just referencing the item directly.

### DIFF
--- a/roles/rac-gi-setup/tasks/rac-asm-create.yml
+++ b/roles/rac-gi-setup/tasks/rac-asm-create.yml
@@ -48,7 +48,7 @@
     LD_LIBRARY_PATH: "{{ grid_home }}/lib:${LD_LIBRARY_PATH}"
   when:
     - asm_disk_management == "asmlib"
-    - created_dg is not search('{{ item.diskgroup }}')
+    - created_dg is not search(item.diskgroup)
   with_items:
     - "{{ asm_disks }}"
   register: create_dg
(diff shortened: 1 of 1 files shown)


jcnars:
Can you please provide steps to reproduce the error message in the initial comment? Thanks

jcnars:
1) Can we pls have the o/p generated :
* with the fix proposed?
2) There are multiple references in the file where this may affect the logic used. To ensure we don't introduce any unwanted effects and to understand how these variables expand: Can you pls add debug tasks and post the output of these variables (if needed iterating thru'):
*  created_dg 
*  item.diskgroup 
* '{{ item.diskgroup }}'
* {{ item.diskgroup }} 

Thanks

mfielding:
1. With the fix, we no longer see a warning, just a regular OK:

```
TASK [rac-gi-setup : rac-asm-create | Check available ASM disk groups] *********
ok: [toolkit-ol8]
```

2. Added debug.print output as requested, and did a sample run:
 
https://gist.github.com/mfielding/bc3d73ed9fb854c292ec6423cebb3adf

jcnars (approved):

PR #141, opened 2023-01-24
Resolve NullPointerException for 19.3 OL8 install

mfielding:
The cluster verification utility run in the grid install can error out with a NullPointerException with Oracle Linux 8 and early Oracle 19.3 software releases.  It occurs because the CVU bundled with early 19.3 releases pre-dates the availability of Oracle Linux 8, and doesn't take its configuration into account.  MOS note 2878100.1 expands on the issue a bit.

```
TASK [rac-gi-setup : rac-gi-install | Run installer] ***************************
fatal: [toolkit-ol8]: FAILED! => {"changed": true, "cmd": ["/u01/app/19.3.0/grid/gridSetup.sh", "-silent", "-responseFile", "/u01/app/19.3.0/grid/gridsetup.rsp", "-J-Doracle.install.mgmtDB=false", "-J-Doracle.install.mgmtDB.CDB=false", "-J-Doracle.install.crs.enableRemoteGIMR=false", "-ignorePrereqFailure"], "delta": "0:00:01.956342", "end": "2023-01-24 16:59:54.634864", "failed_when_result": true, "msg": "non-zero return code", "rc": 255, "start": "2023-01-24 16:59:52.678522", "stderr": "", "stderr_lines": [], "stdout": "Launching Oracle Grid Infrastructure Setup Wizard...\n\n[WARNING] [INS-08101] Unexpected error while executing the action at state: 'supportedOSCheck'\n   CAUSE: No additional information available.\n   ACTION: Contact Oracle Support Services or refer to the software manual.\n   SUMMARY:\n       - java.lang.NullPointerException\nMoved the install session logs to:\n /u01/app/oraInventory/logs/GridSetupActions2023-01-24_04-59-52PM", "stdout_lines": ["Launching Oracle Grid Infrastructure Setup Wizard...", "", "[WARNING] [INS-08101] Unexpected error while executing the action at state: 'supportedOSCheck'", "   CAUSE: No additional information available.", "   ACTION: Contact Oracle Support Services or refer to the software manual.", "   SUMMARY:", "       - java.lang.NullPointerException", "Moved the install session logs to:", " /u01/app/oraInventory/logs/GridSetupActions2023-01-24_04-59-52PM"]}
```

[...]

PR #144, opened 2023-01-26
Fix data types in version comparison

mfielding:
The ansible fact `ansible_distribution_major_version` is actually a string, not a number.  The implicit conversion succeeds in CentOS 7, but with Debian 11 (which has Ansible 2.10 rather than 2.9), I get an error:

```
TASK [rac-gi-setup : rac-gi-install | Patch OL8 CVU issue MOS note 2878100.1] ***
fatal: [database-vm]: FAILED! => {"msg": "The conditional check 'ansible_distribution_major_version >= 8' failed. The error was: Unexpected templating type error occurred on ({% if ansible_distribution_major_version >= 8 %} True {% else %} False {% endif %}): '>=' not supported between instances of 'AnsibleUnsafeText' and 'int'\n\nThe error appears to be in '/home/mfielding_google_com/bms-toolkit/roles/rac-gi-setup/tasks/rac-gi-install.yml': line 97, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: rac-gi-install | Patch OL8 CVU issue MOS note 2878100.1\n  ^ here\n"}
```

Quoting the comparator to consider it as a string too.  Making the same change to the database check as well.

After the change:

```
TASK [rac-gi-setup : rac-gi-install | Patch OL8 CVU issue MOS note 2878100.1] ***
changed: [database-vm]
...

### DIFF
--- a/roles/rac-db-setup/tasks/rac-db-install.yml
+++ b/roles/rac-db-setup/tasks/rac-db-install.yml
@@ -120,7 +120,7 @@
     line: "CV_ASSUME_DISTID=OEL7.8"
   when:
     - ansible_distribution == 'OracleLinux'
-    - ansible_distribution_major_version >= 8
+    - ansible_distribution_major_version >= '8'
     - oracle_ver_base == '19.3'
 
 - name: rac-db-install | Set installer command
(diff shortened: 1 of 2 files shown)


jcnars:
LGTM, also I would imagine errors like these can be caught in the original PR (#141)'s testing.

I wonder if getting a reproducible test case for the reviewer to reproduce a proposed fix will help with that.

jcnars (approved):

PR #154, opened 2023-11-16
add support for nvme disks

alex-basinov:
This PR adds support for NVME disks


alex-basinov:
Hi,

I've updated the code and re-ran the automated test, and the test has now been successfully completed.

The following are links to the test runs: 

- [A test run with NVME disks specified in the format of /dev/nvme0n?](https://gist.github.com/alex-basinov/2a6af3a123431848a29ee92d7df1e297)
- [A test run with NVME disks specified in the format of /dev/disk/by-id/<disk_id>](https://gist.github.com/alex-basinov/13c9c3e1d72b8963d0379c239f3e617e)
- [A test run with SCSI disks specified in the format of /dev/sd?](https://gist.github.com/alex-basinov/ca5c900d2669878192a92471526ccc4b)

Please note that the last 2 test runs progressed only until the point where Ansible tried to download a zip file from the GCS bucket. Ansible fails at that stage due to missing zip files in the bucket, but I think we can ignore the remaining tasks since disk partitioning, swap creation and disk mounting happens prior to that step. That said, please let me know if you'd prefer  to run the install-oracle.sh to its completion.

mfielding (approved):
Looks good;  my tests are succeeding too.

PR #162, opened 2024-03-08
Add support for RHEL9 and OL9

alex-basinov:
b/308784280

This PR adds support for RHEL9 and OL9 OSes.

The changes included are as follows:

1. Created a copy of the oracle_required_rpms_el8 list that excludes RPM packages that are no longer available in RHEL9.
2. Set vm.hugetlb_shm_group to the "oinstall" group's ID, enabling its members to create shared memory segments and mitigating the occurrence of "ORA-27125: unable to create shared memory segment" error during the database creation.

PR #163, opened 2024-03-13
Add support for RHEL9 and OL9 (#162)

alex-basinov:
Integrating changes from the "toolkit-ol9" branch into the "master" branch.

### DIFF
--- a/roles/ora-host/tasks/main.yml
+++ b/roles/ora-host/tasks/main.yml
@@ -87,6 +87,13 @@
   when:
     - install_os_packages|bool and ansible_distribution == 'OracleLinux'
 
+- name: Stat OL9 repo
+  stat:
+    path: /etc/yum.repos.d/oracle-linux-ol9.repo
+  register: oraclelinux_repo_oel9
+  when:
+    - install_os_packages|bool and ansible_distribution == 'OracleLinux'
+
 - name: Install Oracle required packages (base/non-rhui config) for OEL7
   yum:
     name: "{{ oracle_required_rpms }}"
@@ -107,6 +114,18 @@
     - redhat_repo.stat.exists|bool or rh_cloud_repo.stat.exists|bool
   tags: os-packages
 
+- name: Install Oracle required packages (base/non-rhui config) for RHEL9
+  yum:
+    name: "{{ oracle_required_rpms_el9 }}"
+    state: present
+    lock_timeout: 180
+  when:
+    - install_os_packages|bool
+    - ansible_distribution == 'RedHat'
+    - ansible_distribution_major_version == '9'
+    - rh_cloud_repo.stat.exists|bool
+  tags: os-packages
+
 - name: Install Oracle required packages (base/non-rhui config) for OEL8
   yum:
     name: "{{ oracle_required_rpms_el8 }}"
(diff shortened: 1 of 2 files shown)


mfielding (approved):

roles/ora-host/tasks/main.yml
```diff
@@ -178,6 +209,31 @@
...
+- name: Set hugetlb_shm_group to the ID of the oinstall group
+  sysctl:
+    name: vm.hugetlb_shm_group
+    value: "{{ oinstall_group_id.stdout }}"
+    state: present
+    reload: true
+  become: true
+  when:
+    - ansible_distribution == 'RedHat'
+    - ansible_distribution_major_version == '9'
```
mfielding, line 235: Is this actually RHEL9 specific?  https://access.redhat.com/solutions/1610713 lists is as being hte case with RHEL6+. 
And, for that matter, shouldn't we expect the same in Oracle Linux too?

roles/ora-host/tasks/main.yml
```diff
@@ -178,6 +209,31 @@
     - "{{ sysctl_entries }}"
   tags: sysctl
 
+- name: Get the ID of the oinstall group
```
mfielding, line 212: The oinstall gid is actually defined in roles/common/defaults/main.yml would it be cleaner to get it there instead of shelling out?

alex-basinov: I think it would not be cleaner because oracle_groups that is defined in roles/common/defaults/main.yml file is a list containing a bunch of dictionaries. I would have to use a loop and a conditional to find the right item (oinstall group) in the list in order to reference it here. 

As an alternative I could, of course, reference the right item by its index, e.g.: {{ oracle_groups[0].gid }}. This is less flexible and assumes that oinstall is always the first item in the list. The looping or conditional method is more robust if the order of the list might change.  if a separate task is required solely to obtain the group's value, it doesn't significantly differ from simply using the shelling out method.

roles/ora-host/tasks/main.yml
alex-basinov: The ORA-27125 error seems to only impact RH9. When I used the bms-toolkit, cloned from the toolkit-ol9 branch, on RH8, I didn't encounter this error, and the installation finished successfully. See the installation log here: https://gist.github.com/alex-basinov/f8c5ee903f24b775cc988ef6ba0f0fd8
It also didn't fail on OL9: https://gist.github.com/alex-basinov/bda41a2af7f2b12fb48fe2320fc36809

Note that I haven't tested on RH6 and RH7. However, considering that this error has not appeared previously, it's reasonable to conclude that RH9 is the only version impacted.

roles/ora-host/tasks/main.yml
mfielding: Using https://www.redhat.com/sysadmin/ansible-jinja-lists-dictionaries as an example, how about something like (untested) "{{ oracle_groups | selectattr('group', 'match', 'oinstall') | map(attribute='gid') }}" ?

roles/ora-host/tasks/main.yml
mfielding: Fair enough;  we can modify later if we get reports of issues.

roles/ora-host/tasks/main.yml
alex-basinov: Great idea. It hadn't occurred to me to use a Jinja2 expression for extracting GID. 

Filed https://github.com/google/bms-toolkit/pull/164 for this.

mfielding (approved):

PR #165, opened 2024-04-15
Handle ASM diskgroups with multiple disks

alex-basinov:
b/332929223

This PR resolves an issue where populate-asm-disks.yml doesn't correctly manage ASM disk groups containing multiple disks.

I have run tests on both a GCE instance and a BMX machine, each configured with multiple disks:

[Installation log from the GCE instance](https://gist.github.com/alex-basinov/5e919a53b67652650be0e625518ac0a7)
[Installation log from the BMX machine](https://gist.github.com/alex-basinov/c6b8e461f8efd1a94b4d202f2c85c72d)

### DIFF
--- a/roles/common/tasks/populate-asm-disks.yml
+++ b/roles/common/tasks/populate-asm-disks.yml
@@ -6,37 +6,57 @@
       - devices
   tags: populate-asm-disks
 
-- name: Resolve the symlinks
-  ansible.builtin.shell:
-    cmd: "realpath {{ item }}"
-  loop: "{{ asm_disks | map(attribute='disks') | flatten | map(attribute='blk_device') }}"
-  register: realpath_output
+- name: Determine real paths of block devices
+  ansible.builtin.shell: "readlink -f {{ item.1.blk_device }}"
+  with_subelements:
+   - "{{ asm_disks }}"
+   - disks
+  register: real_paths
+  changed_when: false
   tags: populate-asm-disks
 
-- name: Create a copy of the asm_disks with the blk_device set to the canonical path
+    
+- name: Update asm_disks with real paths
   set_fact:
-    asm_disks_normalized: "{{ asm_disks_normalized | default([]) + [item.1 | combine({'disks': item.1.disks | map('combine', {'blk_device': item.0.stdout}) | list })] }}"
-  with_together:
-    - "{{ realpath_output.results }}"
-    - "{{ asm_disks }}"
-  when: item.0.rc == 0 and item.1 is not none
-  tags: populate-asm-disks
-
-- name: Update asm_disks with partition IDs
-  set_fact:
-    asm_disks_with_partitions: "{{ asm_disks_with_partitions | default([]) + [{'diskgroup': item.diskgroup, 'disks': item.disks | map('combine', {'first_partition_id': first_partition_id}) | list}] }}"
-  with_items: "{{ asm_disks_normalized }}"
+    asm_disks_normalized: "{{ asm_disks_updated | from_yaml }}"
   vars:
-    first_partition_id: "{{'/dev/' + ansible_facts.devices[item.disks | map(attribute='blk_device') | join() | basename ].partitions.keys() | first }}"
-  loop_control:
-    label: "{{ item.diskgroup }}"
-  when:
-    - asm_disks_normalized is defined
+    asm_disks_updated: >
+      {% set disks = asm_disks | json_query('[].{diskgroup: diskgroup, disks: disks}') %}
+      {% for disk_group in disks %}
+        {% for disk in disk_group.disks %}
+          {% set real_path = real_paths.results | selectattr('item.1.blk_device', 'equalto', disk.blk_device) | map(attribute='stdout') | list %}
+          {% if real_path %}
+            {% set _ = disk.update({'real_path': real_path | first}) %}
+          {% endif %}
+        {% endfor %}
+      {% endfor %}
+      {{ disks }}
+
+
+- name: Add partition IDs to asm_disks
+  set_fact:
+    asm_disks: "{{ asm_disks_partitions | from_yaml }}"
+  vars:
+    asm_disks_partitions: >
+      {% set updated_disks = [] %}
+      {% for disk_group in asm_disks_normalized %}
+        {% set updated_group_disks = [] %}
+        {% for disk in disk_group.disks %}
+          {% set real_device = disk.real_path | regex_replace('^/dev/', '') %}
+          {% set device_facts = ansible_facts.devices[real_device] if real_device in ansible_facts.devices else {} %}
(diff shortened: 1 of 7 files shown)


mfielding (commented):
Since I'm a bit fuzzy on context, trying to go back to first principles.  We're attempting to populate the first_partition_id of various types of devices, realizing that there are different naming structures in Linux.

Developing a few testcases:

/dev/sda -> /dev/sda1
/dev/mapper/3600a098038314636312451686d4f356a > /dev/mapper/3600a098038314636312451686d4f356ap1
/dev/nvme0n1 -> /dev/neme0n1p1
/dev/disk/by-id/google-data1 -> /dev/disk/by-id/google-data1-part1
/dev/disk/by-id/wwn-0x600a0980383149754d3f54636a73592d -> /dev/disk/by-id/wwn-0x600a0980383149754d3f54636a73592d-part1

In https://gist.github.com/alex-basinov/5e919a53b67652650be0e625518ac0a7 (GCP), I see:

```
changed: [abasinov-rh8] => (item=[{'diskgroup': 'DATA'}, {'blk_device': '/dev/disk/by-id/google-data1', 'name': 'DATA_1', 'real_path': '/dev/sdc', 'first_partition_id': '/dev/sdc1'}])
```

The installation succeeds, but I'm not sure this is as good as `/dev/disk/by-id/google-data1-part1`;  /dev/sdX assignments are not guaranteed to be always the same;  a cloned VM might be one example.
(I don't see a NVME case here, but I suspect it would be similar)

In https://gist.github.com/alex-basinov/c6b8e461f8efd1a94b4d202f2c85c72d (BMS), I see:

```
changed: [at-2944665-svr003] => (item=[{'diskgroup': 'DATA'}, {'blk_device': '/dev/mapper/3600a098038314344373f50416c526770', 'name': 'DATA1', 'real_path': '/dev/dm-6', 'first_partition_id': 'No partition'}])
```

Similarly, `/dev/dm-6` isn't guaranteed to always be the same as `/dev/mapper/3600a098038314344373f50416c526770` over time if disks are added/removed and especially if reimaged.

Or is there a case I'm forgetting where the ASM will refuse to use /dev/disk/by-id?  Or there another mechanism to retain device persistence when device bindings may change?

mfielding:
Any chance of sample output with 75c632a applied?

roles/common/tasks/populate-asm-disks.yml
```diff
@@ -1,62 +1,43 @@
...
+  ansible.builtin.shell: >
+    set -o pipefail;
+    case "{{ item.blk_device }}" in
+      /dev/disk/by-id/*)
+        part_path="{{ item.blk_device }}"-part1;
+        ;;
+      /dev/mapper/*)
+        part_path="{{ item.blk_device }}"p1;
+        ;;
+      /dev/*)
```
mfielding, line 11: What's the best way to handle a path that _doesn't_ start with /dev?

alex-basinov: Could you please clarify why we need to handle this? Aren't block devices typically located in the /dev directory?

roles/common/tasks/populate-asm-disks.yml
```diff
@@ -1,62 +1,43 @@
...
-      - devices
+- name: Resolve blk_device paths to first partition paths if they exist
+  ansible.builtin.shell: >
+    set -o pipefail;
+    case "{{ item.blk_device }}" in
+      /dev/disk/by-id/*)
+        part_path="{{ item.blk_device }}"-part1;
+        ;;
+      /dev/mapper/*)
+        part_path="{{ item.blk_device }}"p1;
```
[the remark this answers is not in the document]
alex-basinov: Not necessarily. Those devices are likely logical volumes managed by LVM and they don't need partitioning like traditional block devices. Since they are abstract volumes, they can be resized or relocated without the need to repartition the physical disks. Therefore, we don't need to track their partition paths.
I've updated the code to skip creating first_partition_id field for devices in /dev/mapper folder. PTAL ?

alex-basinov:
1. [installation log from a GCE instance running RH8](https://gist.github.com/alex-basinov/d3240c8d97f8ff913577c7cfbe500b7f)
2. [installation log from a BMS machine running OL8](https://gist.github.com/alex-basinov/fc3caf2440c706ab23c5ba50bcfb6fa6)
3. [installation log from a C3 GCE instance with nvme disks](https://gist.github.com/alex-basinov/e5f943beb19b1c4e316dc11945fba1bd)
4. [installation log from a GCE instance using swap-blk-device flag](https://gist.github.com/alex-basinov/2a855fe979e36a5f7b004858ea52c100
)

roles/common/tasks/populate-asm-disks.yml
mfielding: From a user perspective, what would be the best way to handle a blk_device entry that doesn't start with dev?  Handle it like /dev/mapper and don't set a part_path?  Try and use the default readlink?  Or output a fatal error?

AFAICT, the implicit default case we have now is like /dev/mapper:  part_path is never set, so implicitly empty.

roles/common/tasks/populate-asm-disks.yml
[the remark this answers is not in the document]
mfielding: I'd say sorta-kinda;  although it has been a convention here in the past, the presence of "mapper" in the device name doesn't necessarily indicate a LVM device.  LVM devices can be referenced as /dev/vg_name, and device-mapper can abstract non-LVM devices.  In the bms-toolkit case, actually the most commoon case is for persistent binding of SAN volumes by WWID.

It brings us back to the original question:  when should we, or should we not, partition disks.  I personally hate disk partitions:  they add complexity, risk I/O misalignment and I/O amplification, complicate resizing, and add the reuqirement for a partiion scan before mounting.  However they do add a form of a platform-independent disk label, and most Oracle shops do use them.  And, officially, Oracle recommends their use for ASM: https://docs.oracle.com/en/database/oracle/oracle-database/19/ladbi/creating-das-or-san-disk-partitions-for-oracle-automatic-storage-management.html.  It's worth noting that they recommend only a single whole-disk partition, so it's not partitioning as in traditional sense, but just a way of labelling the disk.

As I'm sure you've seen in creating the testscases though, it's time-consuming and error-prone to have so many different cases.  What do you think about either always partitioning or never partitioning, for consistency?

mfielding:
Looking at a sample system I see even more edge cases:  `/dev/disk/disk/by-id/dm-name-3600a0980383149754d3f54636a73592d1` not -part1, `/dev/disk/by-partlabel` which shouldn't be appended to, /dev/block/252:0 needs a separate lookup, as does /dev/dm-2.  But these issues exist already, so no worse off here.  So let's submit this to unblock the multiple disk cases, but still keep looking at how to reliably handle partitions.

mfielding (approved):

PR #172, opened 2025-01-20
Add standard .gitattributes file and a more comprehensive .gitignore file to minimize changes of accidentally committing files that should be excluded from the repository

simonpane:
## Change Description:

Add common `.gitignore` and `.gitattributes` entries as per standard practices.

## Solution Overview:

Prevents accidental developer commits of superfluous directories such as `.vscode` (wasn't available at the time of the toolkit's creation) and other common files not currently excluded such as `typescript` or ssh key files.

Ensure proper file type storage (i.e. `.png` files are non-text) EOL characters.

## Test Commands:

N/A as only Git repo metadata is being updated.

## Expected Results:

N/A as only Git repo metadata is being updated.


.gitignore
```diff
@@ -1,9 +1,31 @@
...
+**/*.tmp
+**/nohup.out
+**/typescript
+**/vault.yml
 hosts.yml
-.terraform
 inventory
-logs/*.log
 inventory_files/*
+logs/*.log
```
mfielding, line 27: nit:  consider ignoring all of logs/

.gitignore
```diff
@@ -1,9 +1,31 @@
...
+**/*.swp
+**/*.tmp
+**/nohup.out
+**/typescript
+**/vault.yml
 hosts.yml
-.terraform
 inventory
-logs/*.log
 inventory_files/*
```
mfielding, line 26: suggestion: inventory_files/ for consistency;  we don't include this directory currently.

simonpane: @mfielding , just to ensure I understand:  `inventory_files/*` is in the file currently and this PR isn't changing that.  And as per the previous comment, I will add `logs/*.log` back - it was a mistake to remove.

So what exactly are you suggesting here?  thanks

simonpane: Updated accordingly now @mfielding .  Thanks.

mfielding (approved):

PR #173, opened 2025-01-22
Ensure that the sshd parameter ServerAliveInterval is set to avoid connection timeout issues

simonpane:
## Change Description:

Fix issue causing toolkit to "hang" against Oracle Linux Cloud images (https://yum.oracle.com/oracle-linux-templates.html).

## Solution Overview:

Toolkit would "hang" on long running steps when the target server is an Oracle Linux 7 or 8 image created from the [Oracle Linux Cloud Images](https://yum.oracle.com/oracle-linux-templates.html).

Determined that issue was due to differences in the **sshd** configuration parameter `ClientAliveInterval`.

In the RHEL 7/8 cloud images, sshd is configured with `ClientAliveInterval 420`. However, with Oracle Linux 7/8 it is commented out `#ClientAliveInterval 0`.

Commenting out the `ClientAliveInterval` parameter results in it defaulting to a value of `0` meaning client alive messages are disabled. Thus causing the observed problem where the long-running task runs and completes on the target, the target Python3 process ends, yet the Ansible foreground process on the Ansible control node remains alive and waiting for a response. And consequently appears as "hung".

Solution is for the toolkit to verify the `ClientAliveInterval` parameter, set it according to the toolkit specified (Ansible variable) value, and if changed, restart the `sshd` daemon and importantly, also reset the Ansible connection.

Change includes:

- Adding `sshd restart` and `reset_connection` tasks in the `roles/common/handlers/main.yml` handler task file, as such handlers could be required by other roles with future toolkit enhancements and bug fixes.
  - **IMPORTANT NOTE:** Handlers are processed in task file order, not call order, and consequently the `reset_connection` task must appear after the `sshd restart` task.
  - Handlers are only called on task change using the `notify` parameter.
  - Associated task it not limited to OL7/8 - will check, and if required, set the parameter on any supported target OS.
- Adding the `ClientAliveInterval` setting as a new `sshd_params` array of dictionaries. Allowing for future flexibility where other parameter values can be added - and thus also changed by the corresponding task - as the need arises.
  - Set initial value to `ClientAliveInterval 420` to match the RHEL 7/8 standard value.
- Minor task file formatting updates and fixes (i.e. YAML indentation issues).

### Test Prep:

Provision Oracle Linux 7 or 8 target using image imported into Google Cloud from an [Oracle Linux Cloud Images](https://yum.oracle.com/oracle-linux-templates.html) download.

Required `asm_disk_config.json` file content:

```json
[
  {
    "diskgroup": "DATA",
    "disks": [
      {
        "blk_device": "/dev/disk/by-id/google-oracle-asm-1",
        "name": "DATA1"
      }
    ]
  },
  {
    "diskgroup": "RECO",
    "disks": [
      {
        "blk_device": "/dev/disk/by-id/google-oracle-asm-2",
        "name": "RECO1"
      }
    ]
  }
]
```

Required `data_mounts_config.json` file content:

```json
[
  {
    "purpose": "software",
    "blk_device": "/dev/disk/by-id/google-oracle-disk-1",
    "name": "u01",
    "fstype": "xfs",
    "mount_point": "/u01",
    "mount_opts": "nofail"
  }
]
```

Enter the appropriate IP address for the target database server:

```bash
export INSTANCE_IP_ADDR=10.2.80.7
```

### Test details:

```bash
echo -e "Start date: $(date)\n\n"

./install-oracle.sh \
  --ora-version 19 \
  --backup-dest "+RECO" \
  --allow-install-on-vm \
  --no-patch

echo -e "\n\nEnd date: $(date)\n"
```

## Expected Results:

1. Scripts complete successfully without any indefinate hangs or failed tasks against an Oracle Linux target.

1. Oracle database was properly created and is available/usable.

1. The **sshd** client alive messages value was set according to the Ansible parameter specified value: verify using `sudo grep ClientAliveInterval /etc/ssh/sshd_config`.


simonpane:
Gist showing logfile from full test run, including tasks changing the `ClientAliveInterval` value on the target and running of the `restart_ssh` handler:

- [full-test-output-20250122_113728.txt](https://gist.github.com/simonpane/9a765d3648309eb2ae9aaa2fa4113c57)

mfielding:
The test failure looks like https://github.com/ansible/ansible/issues/50306, fixed in somewhere in the Ansible 2.11 cycle (https://github.com/ansible/ansible/pull/73708 https://www.gitclear.com/open_repos/ansible/ansible/release/stable-2.11-branchpoint?page=42).  We currently support [2.8+](https://github.com/google/bms-toolkit/blob/5774f7dffed530b6076fc1b00b2bc152cee717d3/config-db.yml#L21), and the test env uses 2.9.

A nice catch from the E2E testing.  I never thought about ansible-version testcase coverage.

Do we require a new version of Ansible?  Or insert into the Ansible inventory before we start making SSH connections as per https://gist.github.com/df-a/80a21d8162fefbbca130bb50701d85f5?

simonpane:
> Do we require a new version of Ansible? Or insert into the Ansible inventory before we start making SSH connections as per https://gist.github.com/df-a/80a21d8162fefbbca130bb50701d85f5?

Ansible version support matrix:
* https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix

simonpane:
Good suggestions @mfielding .  I agree that we should try to minimize changes to the target server and keep changes toolkit specific if at all possible.

Changing the parameter in the `ansible.cfg` proved unsuccessful due to the toolkit's usage of the `INSTANCE_SSH_EXTRA_ARGS` parameter.  With both set, problem was not resolved.
 
Consequently, revoked the previous commits of the PR and implemented a narrower scope, and toolkit specific solution, only changing the toolkit Ansible `INSTANCE_SSH_EXTRA_ARGS` configuration parameter instead of the target server's sshd configuration.

If a future change removes the shell scripts, this and other ssh settings will nee to be accounted for at at that time.

The `test bms-toolkit-install` job has been re-run and has succeeded.  Therefore, please review and approve if this change is acceptable to you @mfielding .

Thanks

mfielding (approved):

PR #182, opened 2025-01-26
Remove allow_install_on_vm option from the install_oracle.sh script command line parameters

gorjant-pythian:
## Change Description:

Remove the option allow_install_on_vm from the command line parameters.
The option allow_install_on_vm becomes obsolete as we will always use VMs.

## Solution Overview:

* Deleted the  variable from check-instance.yml
* Removed the  option from install-oracle.sh script
* Simplified platform verification logic in the playbook

## Test Commands:

```bash
 ./install-oracle.sh \
  --instance-hostname gorjantodorovski-db-rhel8-vf3a \
  --ora-edition EE \
  --ora-version 19.3.0.0.0 \
  --backup-dest "+RECO" 
```


install-oracle.sh
```diff
@@ -240,7 +240,7 @@ GETOPT_OPTIONAL="$GETOPT_OPTIONAL,backup-redundancy:,archive-redundancy:,archive
 GETOPT_OPTIONAL="$GETOPT_OPTIONAL,backup-start-hour:,backup-start-min:,archive-backup-min:,backup-script-location:,backup-log-location:"
 GETOPT_OPTIONAL="$GETOPT_OPTIONAL,ora-swlib-type:,ora-swlib-path:,ora-swlib-credentials:,instance-ip-addr:,primary-ip-addr:,instance-ssh-user:"
 GETOPT_OPTIONAL="$GETOPT_OPTIONAL,instance-ssh-key:,instance-hostname:,ntp-pref:,inventory-file:,compatible-rdbms:,instance-ssh-extra-args:"
-GETOPT_OPTIONAL="$GETOPT_OPTIONAL,help,validate,check-instance,prep-host,install-sw,config-db,debug,allow-install-on-vm,skip-database-config,swap-blk-device:"
```
mfielding, line 243: There may be scripts and other code that sets this option;  I'm not sure we want to break them without notice.  What do you think about leaving the option in place, but making it a noop?

mfielding: Perhaps some years in the future we can eventually remove it.

mfielding (commented):
Thanks Gorjan.  Two other quick notes:
- Can you please sign the Google CLA?  There's a note in the failed checks on how to do this.
- Can I ask you to add a link to a gist with the output of your test run?

gorjant-pythian:
[gist test](https://gist.github.com/gorjant-pythian/26863203f7c7198b323e8e4270d42858)

install-oracle.sh
gorjant-pythian: I changed back to allow the option.

AlexBasinov:
Thank you Gorjan! 
Could we also remove the reference to this --allow_install_on_vm  flag from the [user-guide.md](https://github.com/google/oracle-toolkit/blob/5420472112c5da17c5d540b52d9345a240fc361b/docs/user-guide.md?plain=1#L2430) as well?

gorjant-pythian:
Yes, the reference can be removed.

> Message ID: ***@***.***>
>

-- 


--

mfielding:
> Yes, the reference can be removed.
> […](#)
> Message ID: ***@***.***>
> --
> --

@gorjant-pythian can you please add the user guide fix PR?  I think there's just one, in the free edition section.

mfielding (changes requested):
Looks good;  please add the user guide update and we should be good to go.

gorjant-pythian:
@mfielding Please note that the reference in the users_guide.md for allow_install_on_vm is only in this example output, which is not present in the document in my branch, but I think it probably not a good idea to alter production of commands:

`Running Ansible playbook: ansible-playbook -i ./inventory_files/inventory_10.2.83.197_FREE  -e allow_install_on_vm=true  check-instance.yml`

simonpane:
Hi @mfielding , regarding https://github.com/google/oracle-toolkit/pull/182#pullrequestreview-2585134458 ,  I have updated the 23ai Free Edition example in the documentation with a new run that doesn't use the option.  So it no longer appears in the .md file.  Please see: https://github.com/google/oracle-toolkit/pull/182/commits/c584418b0442359080b5599d71cef6b67b9aa7de

simonpane:
@mfielding , I also propose making `--allow-install-on-vm` a true no-op argument in the shell script.  There is no usage in the Ansible at all (with the previous changes in this PR) so there is no need to pass it in to the ansible-playbook command.

Consequently I added the small commits https://github.com/google/oracle-toolkit/pull/182/commits/60b5bd323750700ef23040ab26d6353df5918429 and https://github.com/google/oracle-toolkit/pull/182/commits/a47cd4ceaf20d2e2753d1b1d1dbc23ca26842809 .

Those are small, one-line, and independent commits so revert each or both if you would like.

install-oracle.sh
```diff
@@ -473,7 +473,7 @@ while true; do
     export ANSIBLE_DISPLAY_SKIPPED_HOSTS=true
     ;;
   --allow-install-on-vm)
-    ANSIBLE_PARAMS="${ANSIBLE_PARAMS} -e allow_install_on_vm=true"
+    # No-op argument for backwards compatibility only
```
simonpane, line 476: @mfielding , please see me comment https://github.com/google/oracle-toolkit/pull/182#issuecomment-2627754897 and see if you like this or whether you would like a revert commit?

mfielding: Looks good to me :-)

mfielding (approved):

PR #183, opened 2025-01-26
Set the environment variable CV_ASSUME_DISTID=OEL7 before running the RDBMS and GI installers

gorjant-pythian:
## Change Description:

Installation of Oracle versions 18c and below is not officially supported on RHEL/OL 8. 

## Solution Overview:

This commit introduces an environment variable, 'CV_ASSUME_DISTID,' for both grid and RDBMS installation tasks in Ansible playbooks. If the Oracle version is 18.0 or lower, the variable is set to 'OEL7', ensuring compatibility during the installation process. This change enhances the installation scripts by addressing potential issues related to Oracle version handling.

Additionally, this change adds one more exit code, which is NOT treated as a failure - code 250. The return code indicates that The setup ended in a state that indicates “no further configuration required.” 

## Test command

```bash
./install-oracle.sh \
    --ora-swlib-bucket 
    --backup-dest +RECO \
    --ora-version 18.0.0.0.0 \
    --ora-edition EE \
    --ora-swlib-type gcs \
    --instance-hostname gorjantodorovski-db-rhel8-0ayg \
    --instance-ip-addr 10.2.80.96 \
    --ora-data-mounts data_mounts_config.json \
    --ora-asm-disks asm_disk_config.json \
    --allow-install-on-vm
```

### DIFF
--- a/roles/gi-setup/tasks/gi-install.yml
+++ b/roles/gi-setup/tasks/gi-install.yml
@@ -162,16 +162,20 @@
     msg: "Using installer cmd: {{ installer_command }}"
   tags: gi-setup
 
+- name: gi-install | Set CV_ASSUME_DISTID to OEL7 when installing on RHEL8 (MOS Note: 2878100.1)
+  set_fact:
+    cv_distid: "{{ 'OEL7' if ansible_os_family == 'RedHat' 
+                           and (ansible_distribution_major_version | int) >= 8 
+                           and (oracle_ver_base | float) <= 19.3 
+                           else '' }}"
+  tags: gi-setup
+
 - name: gi-install | Run installer
   become: yes
   become_user: "{{ grid_user }}"
   command: "{{ installer_command }}"
   environment:
-    CV_ASSUME_DISTID: OEL7
-  when:
-    - ansible_os_family == 'RedHat'
-    - ansible_distribution_major_version >= '8'
-    - (oracle_ver_base | float) <= 19.3
+    CV_ASSUME_DISTID: "{{ cv_distid }}"
   register: install_grid_software
   failed_when: "('Successfully Setup Software' not in install_grid_software.stdout) or
                 (install_grid_software.rc not in [0,6,250])"
--- a/roles/gi-setup/tasks/gi-install.yml
+++ b/roles/gi-setup/tasks/gi-install.yml
@@ -164,10 +164,7 @@
 
 - name: gi-install | Set CV_ASSUME_DISTID to OEL7 when installing on RHEL8 (MOS Note: 2878100.1)
   set_fact:
-    cv_distid: "{{ 'OEL7' if ansible_os_family == 'RedHat' 
-                           and (ansible_distribution_major_version | int) >= 8 
-                           and (oracle_ver_base | float) <= 19.3 
-                           else '' }}"
+    cv_distid: "{{ 'OEL7' if ansible_os_family == 'RedHat' and (ansible_distribution_major_version | int) >= 8 and (oracle_ver_base | float) <= 19.3 else '' }}"
   tags: gi-setup
 
 - name: gi-install | Run installer
--- a/roles/gi-setup/tasks/gi-install.yml
+++ b/roles/gi-setup/tasks/gi-install.yml
@@ -139,7 +139,7 @@
 
 - name: gi-install | Set installer command
   set_fact:
-    installer_command: "{{ installer_command }} -silent -responseFile {{ swlib_unzip_path }}/grid_install.rsp {{ rel_patch|default('') }} {{ mgmt_option|default('') }}"
+    installer_command: "{{ installer_command }} -silent -responseFile {{ swlib_unzip_path }}/grid_install.rsp {{ rel_patch | default('') }} {{ mgmt_option | default('') }}"
   tags: gi-setup
 
 - name: gi-install | Update GI OPatch
@@ -157,18 +157,6 @@
   become_user: "{{ grid_user }}"
   tags: gi-setup,opatch
 
-- name: gi-install | Patch OL8 CVU issue MOS note 2878100.1
-  become: yes
-  become_user: "{{ grid_user }}"
-  lineinfile:
-    path: "{{ install_unzip_path }}/cv/admin/cvu_config"
-    regexp: "^CV_ASSUME_DISTID"
-    line: "CV_ASSUME_DISTID=OEL7.8"
-  when:
-    - ansible_os_family == 'RedHat'
-    - ansible_distribution_major_version >= '8'
-    - oracle_ver_base == '19.3'
-
 - name: gi-install | Information
   debug:
     msg: "Using installer cmd: {{ installer_command }}"
@@ -178,6 +166,12 @@
   become: yes
   become_user: "{{ grid_user }}"
   command: "{{ installer_command }}"
+  environment:
+    CV_ASSUME_DISTID: OEL7
+  when:
+    - ansible_os_family == 'RedHat'
+    - ansible_distribution_major_version >= '8'
+    - (oracle_ver_base | float) <= 19.3
   register: install_grid_software
   failed_when: "('Successfully Setup Software' not in install_grid_software.stdout) or
                 (install_grid_software.rc not in [0,6,250])"
@@ -304,7 +298,7 @@
   register: install_grid_plugins
   when: oracle_ver not in ['11.2.0.4.0','12.1.0.2.0']
   failed_when: "(   ('Successfully Setup Software' not in install_grid_plugins.stdout) or
-                    (install_grid_plugins.rc not in [0,6]) )
-                and
-                ('The Installer has detected that there are no config tools to execute for the specified Oracle home' not in install_grid_plugins.stdout)"
+                    (install_grid_plugins.rc not in [0,6,250]) )
+               and
+               ('The Installer has detected that there are no config tools to execute for the specified Oracle home' not in install_grid_plugins.stdout)"
   tags: gi-setup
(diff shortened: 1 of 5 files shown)


roles/gi-setup/tasks/gi-install.yml
```diff
@@ -178,6 +166,12 @@
   become: yes
   become_user: "{{ grid_user }}"
   command: "{{ installer_command }}"
+  environment:
+    CV_ASSUME_DISTID: OEL7
+  when:
+    - ansible_os_family == 'RedHat'
+    - ansible_distribution_major_version >= '8'
```
mfielding, line 173: I think this change might break EL7:  since the when conditions aren't satisfied, it won't run the installer at all.  So I think we still need an extra step, that either sets some sort of variable, or patches cvu_config.

roles/gi-setup/tasks/gi-install.yml
```diff
@@ -157,18 +157,6 @@
   become_user: "{{ grid_user }}"
   tags: gi-setup,opatch
 
-- name: gi-install | Patch OL8 CVU issue MOS note 2878100.1
```
mfielding, line 160: Is there a way we can keep the MOS reference, be it in a comment or elsewhere?  For those not familiar, it's not obvious why we would be setting environment variables that point to a different distribution.

roles/gi-setup/tasks/gi-install.yml
```diff
@@ -304,7 +298,7 @@
   register: install_grid_plugins
   when: oracle_ver not in ['11.2.0.4.0','12.1.0.2.0']
   failed_when: "(   ('Successfully Setup Software' not in install_grid_plugins.stdout) or
-                    (install_grid_plugins.rc not in [0,6]) )
-                and
-                ('The Installer has detected that there are no config tools to execute for the specified Oracle home' not in install_grid_plugins.stdout)"
+                    (install_grid_plugins.rc not in [0,6,250]) )
```
mfielding, line 301: This change looks fine, but offhand, can we add a mention in the PR description that mentions what types of failures we encountered that had this return code?

roles/gi-setup/tasks/gi-install.yml
```diff
@@ -304,7 +298,7 @@
...
+               and
+               ('The Installer has detected that there are no config tools to execute for the specified Oracle home' not in install_grid_plugins.stdout)"
```
mfielding, line 303: I realize that this is pre-existing code, but I want to confirm:  is it indeed safe to continue if the installer reports "no config tools to execute"?  I presume it's because this message only appears after a successful install?

roles/rdbms-setup/tasks/rdbms-install.yml
```diff
@@ -97,45 +97,39 @@
...
-  tags: rdbms-setup
+  command: "{{ db_installer_cmd }}"
+  environment:
+    CV_ASSUME_DISTID: OEL7
+  when:
+    - ansible_os_family == 'RedHat'
+    - ansible_distribution_major_version >= '8'
+    - (oracle_ver_base | float) <= 19.3
+  register: db_install_result
+  failed_when:
```
mfielding, line 121: This change appears to change an "or" to an "and".  I don't think that was intentional.

roles/rdbms-setup/tasks/rdbms-install.yml
```diff
@@ -97,45 +97,39 @@
...
   debug:
     msg:
-      - "{{ install_db_software.cmd }}"
-      - "{{ install_db_software.stdout_lines }}"
-    #verbosity: 1
+      - "{{ db_install_result.cmd }}"
+      - "{{ db_install_result.stdout_lines }}"
   tags: rdbms-setup
-
+
```
mfielding, line 132: nit: extra spaces

mfielding (changes requested):

roles/gi-setup/tasks/gi-install.yml
gorjant-pythian: I have changed the logic to use a dynamic set_fact to conditionally set the environment variable CV_ASSUME_DISTID and then set it during installation without additional conditions.

roles/gi-setup/tasks/gi-install.yml
gorjant-pythian: Addressed by notifying the reason for setting the CV_ASSUME_DISTID variable when setting the fact.

roles/gi-setup/tasks/gi-install.yml
gorjant-pythian: Added comment in the PR description about this.

roles/gi-setup/tasks/gi-install.yml
gorjant-pythian: The message "The Installer has detected that there are no config tools to execute for the specified Oracle home" appears when running Oracle Database or Grid Infrastructure installation in silent mode with configuration assistants that need to run `root.sh` before completion[4].

## Cause
When Oracle Universal Installer (OUI) is running in silent mode and detects that `root.sh` needs to be executed prior to configuration assistants, it will:

1. Skip the configuration assistants during the installation[4]
2. Require manual intervention after the silent installation completes[4]

## Required Actions
To properly complete the installation when encountering this message:

1. Let the silent installation finish
2. Run the `root.sh` script as root user
3. Manually execute the skipped configuration assistants[4]

The configuration assistants can be run after installation using the `configToolAllCommands` utility located in the `ORACLE_HOME\cfgtoollogs` directory[1][9].

Sources
[1] Installing and Configuring Oracle Database Using Response Files https://docs.oracle.com/cd/E18790_01/doc/install.112/e16773/advance.htm
[2] Oracle 19c database silent installation {FATAL] [INS-35954] error https://stackoverflow.com/questions/60853123/oracle-19c-database-silent-installation-fatal-ins-35954-error
[3] Troubleshooting the Oracle Grid Infrastructure Installation Process https://dba86.com/docs/oracle/11.2/install.112/e48194/trouble.htm
[4] 2 Using Oracle Universal Installer https://docs.oracle.com/cd/B13789_01/em.101/b12140/2_oui_using.htm
[5] 1 Installation and Configuration Issues https://docs.oracle.com/cd/E49269_01/doc.12/e49552/install_configure.htm
[6] Troubleshooting the Oracle Grid Infrastructure Installation Process https://docs.oracle.com/cd/E11882_01/install.112/e48194/trouble.htm
[7] 2 Managing Oracle Home Directories https://docs.oracle.com/cd/E24628_01/em.121/e37799/ch2_manage_oh.htm
[8] Can not find Oracle home / Points of Confusion - Devart Forums https://forums.devart.com/viewtopic.php?t=20681
[9] Troubleshooting the Oracle Grid Infrastructure Installation Process https://docs.oracle.com/database/121/CWWIN/trouble.htm
[10] runInstaller Command - Oracle Help Center https://docs.oracle.com/en/database/oracle/oracle-database/23/ladbi/runinstaller-command-options.html
[11] Troubleshooting the Installation and Setup - Oracle Help Center https://docs.oracle.com/cd/E11882_01/owb.112/e17130/appendix_troubleshooting.htm
[12] INS-32103 Specified Oracle Home user does not match with existing ... https://stackoverflow.com/questions/70868521/oracle-db-client-19c-install-error-ins-32103-specified-oracle-home-user-does

roles/rdbms-setup/tasks/rdbms-install.yml
gorjant-pythian: Addressed by the change in the current decision logic.

roles/rdbms-setup/tasks/rdbms-install.yml
gorjant-pythian: Cleaned up

roles/rdbms-setup/tasks/rdbms-install.yml
mfielding: I think this issue still exists:  we want to fail if there's an error response code _or_ we don't see a success message.

mfielding (changes requested):
Looking good;  just one change.

simonpane:
I agree with @mfielding here https://github.com/google/oracle-toolkit/pull/183#discussion_r1934345752 - @gorjant-pythian , I think that we do want a failure with either:

```
[LINE 126] "'Successfully Setup Software' not in db_install_result.stdout"
```
Or
```
[LINE 127] - "db_install_result.rc not in [0,6,250]"
```

As it stands, when those are two different list elements, they are treated as a logical **AND**.

(Might be other sections with the same construct to fix as well.)

roles/rdbms-setup/tasks/rdbms-install.yml
gorjant-pythian: It should be okay now. I have changed the check to use the "OR" operator for the output message and the return code.

mfielding:
Calling out the test failure https://oss.gprow.dev/view/gs/bmaas-testing-prow-bucket/pr-logs/pull/google_oracle-toolkit/183/oracle-toolkit-install/1885018665259110400

```
TASK [gi-setup : include_tasks] ************************************************
fatal: [g278813163-s366]: FAILED! => {"reason": "We were unable to read either as JSON nor YAML, these are the errors we got from each:\nJSON: Expecting value: line 1 column 1 (char 0)\n\nSyntax Error while loading YAML.\n  mapping values are not allowed in this context\n\nThe error appears to be in '/home/prow/go/src/github.com/google/oracle-toolkit/roles/gi-setup/tasks/gi-install.yml': line 165, column 85, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: gi-install | Set CV_ASSUME_DISTID to OEL7 when installing on RHEL8 (MOS Note: 2878100.1)\n                                                                                    ^ here\n"}
```

mfielding (approved):

PR #186, opened 2025-01-29
Add functionality for removing Free Edition RPM based installs

simonpane:
## Change Description:

Add Oracle Database Free Edition (and notably RPM based installation) cleanup and remove ability to the brute force cleanup workflow.

## Solution Overview:

Since Oracle Database Free Editions are RPM based installations, significant changes to the the brute force cleanup process is required. Including:

- Changing the `cleanup-oracle.sh` shell script to accept the `ORA_EDITION` environment variable or the `--ora-edition` command line switch to specify that the cleanup is for a Free Edition installation.
  - In the future, there is the possibility that a specific Oracle version may be available in both Free and commercial (SE2/EE) editions and therefore relying on the version alone is not sufficient.
- Major additions to the `brute-force-cleanup` role to perform the required steps to remove the RPM, the systemd services, and related files unique to Free Edition.
  - Due to the relatively deep linking of RPM scriplet activities and service files the overall process is to try removing & uninstalling using the standard process and then reverting to more direct removal/deletion commands if necessary (which may be necessary in the case of a broken or corrupted installation).
  - Since the new tasks performing a systemd reload and reset are generic with a high probability of re-usability by other tasks (possibly in the future with additional toolkit enhancements) or Oracle technology changes, these tasks are added to the `common` role as handlers.

Additional changes include:

- General Ansible fixes and enhancements to the `roles/brute-ora-cleanup/tasks/main.yml` task file to align with best practices including: truthy fixes, string quoting, white-space normalization, and task optimizations and idempotency enhancements as possible in YAML and shell scripts.
- Updated `ansible.cfg` to replace tabs with spaces and include `command_warnings = False` to suppress the Ansible 2.9 warnings on using `rpm` through the `shell`/`cmd` module.  This does not suppress other warnings such as when SELinux is disabled.

### Test Prep:

Required `data_mounts_config.json` file content:

```json
[
  {
    "purpose": "software",
    "blk_device": "/dev/disk/by-id/google-oracle-disk-1",
    "name": "u01",
    "fstype": "xfs",
    "mount_point": "/u01",
    "mount_opts": "nofail"
  }
]
```

Enter the appropriate IP address for the target database server:

```bash
export INSTANCE_IP_ADDR=10.2.80.16
```

Install the "latest" version of 23ai Free Edition by default (i.e. by not explicitly stating a specific version):

```bash
./install-oracle.sh \
  --ora-edition FREE \
  --ora-swlib-bucket gs://pythian-gto-oracle-software/free-edition \
  --backup-dest /opt/oracle/fast_recovery_area/FREE \
  --allow-install-on-vm
```

### Test 1: Cleanup against a fully installed, configured, and operational target:

Run cleanup command:

```bash
./cleanup-oracle.sh \
  --ora-edition FREE \
  --ora-version 23.6.0.24.10 \
  --inventory-file inventory_files/inventory_${INSTANCE_IP_ADDR}_FREE \
  --yes-i-am-sure
```

Steps to test resulting target server:

```bash
hostname -I
find /etc | grep ora | grep -Eiv 'linux|fedora|oracle-release|oracle.com|orabackup|RPM-GPG-KEY-oracle'
find /etc | grep -Ei 'database|23c|23ai'
find /usr | grep -Ei 'database|23c|23ai' | grep -Eiv 'google|gcloud|cpython|mime|pip|build-id|firmware|23com'
find /opt | grep -Ei 'database|23c|23ai'
systemctl | grep -Ei 'oracle|free'
grep oracle /etc/passwd
ps -ef | grep oracle
```

### Test 2: Re-execution of cleanup against an already cleaned up target:

Run cleanup command:

```bash
./cleanup-oracle.sh \
  --ora-edition FREE \
  --ora-version 23.6.0.24.10 \
  --inventory-file inventory_files/inventory_10.2.80.63_FREE \
  --yes-i-am-sure
```

Steps to test resulting target server:

```bash
hostname -I
find /etc | grep ora | grep -Eiv 'linux|fedora|oracle-release|oracle.com|orabackup|RPM-GPG-KEY-oracle'
find /etc | grep -Ei 'database|23c|23ai'
find /usr | grep -Ei 'database|23c|23ai' | grep -Eiv 'google|gcloud|cpython|mime|pip'
find /opt | grep -Ei 'database|23c|23ai'
systemctl | grep -Ei 'oracle|free'
grep oracle /etc/passwd
ps -ef | grep oracle
```

## Expected Results:

- Initial run removes all traces of the Oracle database and software from the target server.
- When run against an installed/configured/running database target, all tasks succeed or are skipped - none failed/ignored.
- New steps/tasks are fully idempotent regardless of state of the target database server - no failed/ignored Free Edition specific tasks.

### DIFF
--- a/cleanup-oracle.sh
+++ b/cleanup-oracle.sh
@@ -18,7 +18,7 @@ echo "$0 $@"
 echo
 
 GETOPT_MANDATORY="ora-version:,inventory-file:,yes-i-am-sure"
-GETOPT_OPTIONAL="ora-role-separation:,ora-disk-mgmt:,ora-swlib-path:,ora-staging:,ora-asm-disks:,ora-data-mounts:,help"
+GETOPT_OPTIONAL="ora-edition:,ora-role-separation:,ora-disk-mgmt:,ora-swlib-path:,ora-staging:,ora-asm-disks:,ora-data-mounts:,help"
 GETOPT_LONG="${GETOPT_MANDATORY},${GETOPT_OPTIONAL}"
 GETOPT_SHORT="yh"
 
@@ -27,7 +27,10 @@ VALIDATE=0
 INVENTORY_FILE="${INVENTORY_FILE:-./inventory_files/inventory}"
 
 ORA_VERSION="${ORA_VERSION}"
-ORA_VERSION_PARAM='^(19\.3\.0\.0\.0|18\.0\.0\.0\.0|12\.2\.0\.1\.0|12\.1\.0\.2\.0|11\.2\.0\.4\.0|ALL)$'
+ORA_VERSION_PARAM='^(23\.[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,6}|21\.3\.0\.0\.0|19\.3\.0\.0\.0|18\.0\.0\.0\.0|12\.2\.0\.1\.0|12\.1\.0\.2\.0|11\.2\.0\.4\.0)$'
+
+ORA_EDITION="${ORA_EDITION:-EE}"
+ORA_EDITION_PARAM="^(EE|SE|SE2|FREE)$"
 
 ORA_ROLE_SEPARATION="${ORA_ROLE_SEPARATION:-TRUE}"
 ORA_ROLE_SEPARATION_PARAM="^(TRUE|FALSE)$"
@@ -63,6 +66,7 @@ while true; do
         ;;
     --ora-version)
         ORA_VERSION="$2"
+        if [[ ${ORA_VERSION} = "23" ]]   ; then ORA_VERSION="23.0.0.0.0"; fi
         if [[ ${ORA_VERSION} = "19" ]]   ; then ORA_VERSION="19.3.0.0.0"; fi
         if [[ ${ORA_VERSION} = "18" ]]   ; then ORA_VERSION="18.0.0.0.0"; fi
         if [[ ${ORA_VERSION} = "12" ]]   ; then ORA_VERSION="12.2.0.1.0"; fi
@@ -70,6 +74,10 @@ while true; do
         if [[ ${ORA_VERSION} = "12.1" ]] ; then ORA_VERSION="12.1.0.2.0"; fi
         if [[ ${ORA_VERSION} = "11" ]]   ; then ORA_VERSION="11.2.0.4.0"; fi
         shift;
+        ;;
+    --ora-edition)
+        ORA_EDITION="$(echo "$2" | tr '[:lower:]' '[:upper:]')"
+        shift
         ;;
     --inventory-file)
         INVENTORY_FILE="$2"
@@ -119,9 +127,14 @@ done
 
 shopt -s nocasematch
 
+# Oracle Database free edition parameter overrides
+if [[ "${ORA_EDITION}" = "FREE" && ! "${ORA_VERSION}" =~ ^23\. ]]; then
+    ORA_VERSION="23.0.0.0.0"
+fi
+
 # Mandatory options
 if [ "${ORA_VERSION}" = "" ]; then
-    echo "Please specify the oracle release with --ora-version"
+    echo "Please specify the oracle version with --ora-version"
     exit 2
 fi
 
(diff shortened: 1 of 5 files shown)


simonpane:
Gist files with outputs from test scenarios:

- [oratk-30: Cleanup output from execution against an installed/configured/running target](https://gist.github.com/simonpane/732525c9168c99e2f744ac8333b8ea79)
- [oratk-30: Target server state after cleanup against installed/configured/running target](https://gist.github.com/simonpane/57a859d98e5eee2bc4bc5bd343d48cd5)
- [oratk-30: Cleanup output from execution against an already cleaned target](https://gist.github.com/simonpane/fdbe50a6b1e978e61a495961e918b76b)
- [oratk-30: Regression test 19c install](https://gist.github.com/simonpane/3f08e5c154d797630a365cf0bfe2073c)
- [oratk-30: Regression test cleanup against installed/configured/running 19c target](https://gist.github.com/simonpane/3198990cd7ac08b4bacec8cbd9127f15)

roles/brute-ora-cleanup/tasks/main.yml
```diff
@@ -84,15 +200,15 @@
     - '\\+ASM'
     - /grid/bin
     - /grid/jdk
-    - "-u {{ grid_user }}"
```
mfielding, line 87: Don't we still want to kill processes running as the grid user?

roles/brute-ora-cleanup/tasks/main.yml
```diff
@@ -112,19 +228,19 @@
...
     - "-u {{ oracle_user }}"
-  ignore_errors: yes
+    - "{% if free_edition %}[]{% else %}-u {{ grid_user }}{% endif %}"
```
mfielding, line 232: Can we just add some quoting instead so the ignore_errors works?  An install without privilege separation would also not have a grid user.  (Or maybe even default it to the same as oracle_user).

roles/brute-ora-cleanup/README.md
```diff
@@ -1,15 +1,35 @@
...
+
+```bash
+./cleanup-oracle.sh \
+  --ora-edition FREE \
+  --ora-version 23.6.0.24.10 \
+  --inventory-file inventory_files/inventory_10.2.80.54_FREE \
+  --yes-i-am-sure
+```
+
 ## Items not covered
```
mfielding, line 33: I think these items are now covered, so this section is probably no longer necessary

cleanup-oracle.sh
```diff
@@ -139,6 +152,10 @@ fi
     echo "Incorrect parameter provided for ora-version: $ORA_VERSION"
     exit 1
 }
+[[ ! "$ORA_EDITION" =~ $ORA_EDITION_PARAM ]] && {
```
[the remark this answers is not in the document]
simonpane: @mfielding , this is not a new mandatory parameter, it is an optional parameter with the same default as before.

If you look in [this gist](https://gist.github.com/simonpane/3198990cd7ac08b4bacec8cbd9127f15) from the comment above, you can see that I ran a regression clean-up against a 19c database with the exact same parameters as we would have used before this change and it succeeded and worked the same as before this change.

roles/brute-ora-cleanup/tasks/main.yml
simonpane: Hi @mfielding , a couple of insights on this comment:

1. There is no change to the `ignore_errors` functionality in this change.  It it sill needed to prevent the the playbook from failing if the script is re-run against a target that has already been cleaned.

2. There are numerous errors in various playboooks that need to be fixed for when `role_separation = False`.  I can update this one here but I also have a ticket in the backlog for updating in other places.

roles/brute-ora-cleanup/tasks/main.yml
simonpane: @mfielding , yes we do, but only if required.  To prevent unnecessary errors (allowing it to run through cleanly) when uninstalling Free Edition I made it conditional in the subsequent line.

roles/brute-ora-cleanup/README.md
simonpane: Yes you are correct @mfielding - I have removed that section in https://github.com/google/oracle-toolkit/pull/186/commits/6dc47b43c5d0bd23262789e5ec048f2e76ace74d.

cleanup-oracle.sh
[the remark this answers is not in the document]
mfielding: This is great, thanks!

mfielding (approved):

PR #187, opened 2025-01-29
Cleanup process enhancement: improved if-statement to instead check on `role_separation`

simonpane:
Enhancement to base the conditional check on the `role_separation` variable.

Re-tested execution:

- If there is `role_separation == TRUE` then it will `pkill` for both the **oracle** and **grid** users.  When not (including for Free Edition), it will only `pkill` processes by the **oracle** user and therefore avoid unnecessary failed messages due to trying to kill processes for a user that is not applicable to the installation.

### DIFF
--- a/cleanup-oracle.sh
+++ b/cleanup-oracle.sh
@@ -127,9 +127,11 @@ done
 
 shopt -s nocasematch
 
-# Oracle Database free edition parameter overrides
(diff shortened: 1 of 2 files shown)


simonpane:
Created this PR as an enhancement based on the discussion in comment: https://github.com/google/oracle-toolkit/pull/186#discussion_r1934338265

mfielding (approved):

PR #197, opened 2025-02-10
Fix 19.25 patch name in user guide

mfielding:
The patch metadata is correct, but there's just a placeholder p0000000 in the documentation.  Updating that.


simonpane:
Good catch @mfielding - sorry I missed that originally.

simonpane (approved):
Reviewed & approved

simonpane (approved):

PR #199, opened 2025-02-11
Optimize kernel parameter settings and remove duplicate section from the `hugepages` role

simonpane:
## Change Description:

Improve the inefficient, duplicated, not well written, and noisy kernel parameter setting sections.

## Solution Overview:

The toolkit is setting kernel parameters in two roles: `ora-host` and `hugepages`. In the case of the latter, the Ansible is not well-written, uses complicated and out-dated logic, and includes numerous (arguably noisy) individual debug statements with no verbosity control.

Further, not all of these kernel parameter settings are not directly related to setting Linux huge pages - although they do support aspects of Linux huge pages. Consequently, this change standardizes the kernel parameter settings in the `ora-host` role and removes them from the `hugepages` role. The `ram_pct_used` variable, previously defined under the `hugepages` role is similarly moved to the `ora-host` role.

> **NOTE:** Other changes to the `hugepages` steps and Ansible tasks are not included in this change and will be addressed separately in a subsequent change.
>
> This change is focused on the Linux kernel parameter settings. Other kernel resource limits also outside of the scope of this specific change. One exception to this is the `pam_limits` task which was also duplicated in `hugepages` and `ora-host` - removed from the former and kept in the latter.

Updated logic is based on: MOS Notes:

- [Maximum SHMMAX values for Linux x86 and x86-64 (Doc ID 567506.1)](https://support.oracle.com/epmos/faces/DocContentDisplay?id=567506.1)
- [Relationship Between SGA_TARGET and SHMMAX (Doc ID 1527109.1)](https://support.oracle.com/epmos/faces/DocContentDisplay?id=1527109.1)
- [Health Check Alert: Set SHMMAX greater than or equal to the recommended minimum value (Doc ID 957520.1)](https://support.oracle.com/epmos/faces/DocContentDisplay?id=957520.1)
- [Health Check Alert: Set SHMALL greater than or equal to the recommended minimum value (Doc ID 957445.1)](https://support.oracle.com/epmos/faces/DocContentDisplay?id=957445.1)

Output of the new tasks is concise yet meaningful:

```
TASK [ora-host : kernel_parameters | Capture existing sysctl values] *******************************************
ok: [10.2.80.117]

TASK [ora-host : kernel_parameters | Convert existing sysctl values to a list of dictionaries] *****************
ok: [10.2.80.117]

TASK [ora-host : kernel_parameters | Update the shmall sysctl value (if less than "ram_pct_used" the size of physical memory)] ***
changed: [10.2.80.117]

[...]

PR #206, opened 2025-02-18
Move and optimize Linux hugepages tasks

simonpane:
## Change Description:

Move the tasks from the existing `hugepages` role to a task file under the `ora-host` role and optimize tasks.

## Solution Overview:

A role should ideally be a noun that describes a server, for example "database-server".  Hugepage configuration isn't really a noun describing a server, it's more of a characteristic of an `ora-host` server.

Further, other similar configuration settings (such as kernel parameter settings) are handled under tasks as part of the `ora-host` role.

Consequently, have moved the tasks from the `hugepages` role to a task file under the the `ora-host` role.  Keeping these tasks in a separate task file does make sense as the `main.yml` task file is already excessively large.

Additionally some debug hugepages tasks are overly verbose and not really necessary and consequently have been removed to make the overall playbook output more concise, only displaying useful debug output. 

Finally, having multiple `shell` module tasks based on the same `when:` key value is redundant and consequently the actions of such tasks can be combined.

## Test Results:

Before running toolkit (on managed host / database server target):

```bash
$ grep -E '(AnonHugePage|HugePages_Total)' /proc/meminfo
AnonHugePages:    247808 kB
HugePages_Total:       0
```

Gist [output log](https://gist.github.com/simonpane/b12eb78f4da0ef7b76f5941a252fb39a) from full test run of toolkit.

After running toolkit **and rebooting** the managed host:

```bash
$ grep -E '(AnonHugePage|HugePages_Total)' /proc/meminfo
AnonHugePages:         0 kB
HugePages_Total:    3687
```

Gist [output log](https://gist.github.com/simonpane/88017e52ee9022c3bbc21356ebf27b09) showing that effected section is idempotent (although other sections not related to this change may not be fully idempotent).


roles/ora-host/tasks/hugepages.yml
```diff
@@ -13,51 +13,56 @@
...
+  debug:
     msg: "WARNING: A reboot is required to disable transparent huge pages"
+    verbosity: 0
   when: checkTHPreboot.stdout == "reboot"
+
+# Setting vm.hugetlb_shm_group to the ID of the "oinstall" group to allow its members to create
+# SysV shared memory segment using hugetlb page.
+# By default only the root user has permissions to create shared memory segments.
+# This is needed to prevent the occurrence of "ORA-27125: unable to create shared memory segment" error.
+- name: hugepages | Set hugetlb_shm_group to the ID of the oinstall group
```
mfielding, line 59: Nice catch;  It's also possible for members of the dba and oper groups to start up an instance.  Is there a simple way to handle this?

roles/ora-host/tasks/hugepages.yml
```diff
@@ -13,51 +13,56 @@
...
+  sysctl:
+    name: vm.hugetlb_shm_group
+    value: "{{ oracle_groups | selectattr('group', 'match', 'oinstall') | map(attribute='gid') | first }}"
+    state: present
+    reload: true
+  become: true
+  when:
+    - ansible_distribution == 'RedHat'
+    - ansible_distribution_major_version == '9'
```
mfielding, line 68: https://access.redhat.com/solutions/1610713 suggest that this issue goes back to RHEL5 at least.

roles/ora-host/tasks/hugepages.yml
```diff
@@ -13,51 +13,56 @@
...
+# This is needed to prevent the occurrence of "ORA-27125: unable to create shared memory segment" error.
+- name: hugepages | Set hugetlb_shm_group to the ID of the oinstall group
+  sysctl:
+    name: vm.hugetlb_shm_group
+    value: "{{ oracle_groups | selectattr('group', 'match', 'oinstall') | map(attribute='gid') | first }}"
+    state: present
+    reload: true
+  become: true
+  when:
+    - ansible_distribution == 'RedHat'
```
mfielding, line 67: Non-RedHat distros have this issue too, right?

roles/ora-host/tasks/hugepages.yml
```diff
@@ -13,51 +13,56 @@
...
-- name: get Hugepagesize from memory
+- name: hugepages | get Hugepagesize from memory
   shell: set -o pipefail; ( cat /proc/meminfo | ( grep Hugepagesize || true ) | cut -d " " -f8 ) || true
   register: v_Hugepagesize
   changed_when: false
 
-- name: Update the vm.nr_hugepages sysctl value
+- name: hugepages | Update the vm.nr_hugepages sysctl value
   sysctl:
     name: "vm.nr_hugepages"
```
mfielding, line 23: Not a new issue, but should we also update the cleanup script to remove the hugepages?

roles/ora-host/tasks/hugepages.yml
```diff
@@ -13,51 +13,56 @@
...
+    verbosity: 0
   when: checkTHPreboot.stdout == "reboot"
+
+# Setting vm.hugetlb_shm_group to the ID of the "oinstall" group to allow its members to create
+# SysV shared memory segment using hugetlb page.
+# By default only the root user has permissions to create shared memory segments.
+# This is needed to prevent the occurrence of "ORA-27125: unable to create shared memory segment" error.
+- name: hugepages | Set hugetlb_shm_group to the ID of the oinstall group
+  sysctl:
+    name: vm.hugetlb_shm_group
```
mfielding, line 61: Consider setting this in roles/ora-host/defaults/main.yml, where we set most other sysctl parameters (including hugepages)

roles/ora-host/tasks/kernel_parameters.yml
```diff
@@ -34,14 +34,14 @@
 - name: kernel_parameters | Update the shmall sysctl value (if less than "ram_pct_used" the size of physical memory)
```
mfielding, line 34: Officially, https://docs.oracle.com/en/database/oracle/oracle-database/19/ladbi/minimum-parameter-settings-for-installation.html says that shmmax should be half of physical memory (in bytes), and that shmall should be >= shmmax (but measured in pages).

So, strictly speaking, should we calculate shmmax too?  (Having ram_pct_used at 65% by default makes sense though with today's high-memory hardware)

roles/ora-host/tasks/hugepages.yml
simonpane: Hi @mfielding , this block was actually added by @AlexBasinov a few years ago.  It's only showing up as a change here as I moved it's place.  No change to the substance of the task.  FYI

roles/ora-host/tasks/kernel_parameters.yml
simonpane: Hi @mfielding , it is calculated as per Oracle's guidance in file [roles/ora-host/defaults/main.yml](https://github.com/google/oracle-toolkit/blob/master/roles/ora-host/defaults/main.yml) and as per the MOS Docs in https://github.com/google/oracle-toolkit/pull/199.

Important to that is that `shmmax` doesn't have to match ram_pct_used (65%) or equal the size of the the cumulative SGAs.

roles/ora-host/tasks/hugepages.yml
simonpane: Yes, also found MOS [Doc ID 2491966.1](https://support.oracle.com/epmos/faces/DocContentDisplay?id=2491966.1) that confirms Linux 6+

simonpane: Removed this criteria in commit https://github.com/google/oracle-toolkit/pull/206/commits/e6e1b77a2f3c2183b8ae3dd9fdba2575de319896.

roles/ora-host/tasks/hugepages.yml
simonpane: Yes agreed.  Removed this criteria (and ran a manual test) in commit https://github.com/google/oracle-toolkit/pull/206/commits/e6e1b77a2f3c2183b8ae3dd9fdba2575de319896.

roles/ora-host/tasks/hugepages.yml
simonpane: Good suggestion @mfielding , did exactly that in https://github.com/google/oracle-toolkit/pull/206/commits/e6e1b77a2f3c2183b8ae3dd9fdba2575de319896 (and tested).

roles/ora-host/tasks/hugepages.yml
simonpane: @mfielding , I added a first version of this in commit https://github.com/google/oracle-toolkit/pull/206/commits/ca1f8690d7bcaafa750c5dda8df1a4aab05d95c6.  If you don't want to include it in this change, we can always revert it and handle it separately.  Also, if you want to move those parameters into a `defaults/main.yml` parameter file we could.  Up to you on both of those points.

roles/brute-ora-cleanup/tasks/main.yml
```diff
@@ -456,6 +456,46 @@
   become_user: root
   shell: awk '/^oracle/ {print $1}' /proc/modules | sudo xargs -r modprobe -r
 
+- name: Reset kernel sysctl values to EL defaults
+  sysctl:
+    name: "{{ item.parameter }}"
+    value: "{{ item.value }}"
+    state: present
+    sysctl_set: true
```
mfielding, line 464: In a perfect world, we'd set the parameters to what we think are Red Hat's defaults on the live system, but remove entirely from /etc/sysctl.conf.  But given that the ANsible sysctl module doesn't have an option for that, what's here should be fine.

mfielding (approved):
Looks good to me, but I merged your 1mb fix before this, so will need the conflicts resolved to merge this.

simonpane:
Merge conflicts resolved.

mfielding (approved):

PR #213, opened 2025-02-21
Optimize playbook "readiness" checks into a concise and consistent set of pre_tasks

simonpane:
## Change Description:

Optimize playbook "readiness" checks into a concise, efficient, and consistent set of **pre_tasks**. Using a framework approach that can be expanded on, as required, in the future. Initial PR provides the framework and the minimal set of Ansible control node and managed hosts checks.

With the goal of detecting any issues early and quickly (aka "fail-fast") to prevent later playbook failures. While removing redundant tasks that are duplicated in every playbook (such as checking the Ansible version) and providing a more concise yet meaningful STDOUT.

## Solution Overview:

For all playbooks, there are some initial checks that we can perform proactively and early on in the `pre_tasks` section to ensure that the Ansible control node, and managed hosts are ready.

> **NOTE:** Calling these "_**readiness checks**_" to not overlap with, or cause confusion with, the Oracle database server and and instance "validation" scripts used subsequently.

In all playbooks we should be ensuring certain aspects of the Ansible control node including, at a minimum:

- The Ansible version
- Whether JMESpath is installed (via pip3 or dnf)

Additionally, for most playbooks we also should verify managed host aspects such as:

- OS family, distribution, version
- Chip-set architecture (toolkit may in the future be expanded to support ARM-based chips)
- Minimum memory and swap space to support a database instance
- Privilege escalation ability (which may not be via sudo and may possibly in a future enhancement, be to a superuser other than "root")

Checking these criteria consistently adds overall reliability in playbooks. And checking them early prevents incomplete long running deployments (for example the scenario where the toolkit runs for an extended period of time installing software only to eventually fail when trying to run `root.sh` with privilege escalation during a late stage task).

---

This change implements a flexible readiness check solution based on three variables initially:

1. `control_node_checks`: a boolean (defaults to `false`) to control checks specific to the Ansible control node.
1. `managed_host_checks`: a boolean (defaults to `false`) to control checks specific to the managed hosts (aka "targets").
1. `become_user_check`: a string value of a user to test privilege escalation to (i.e. `root`) via the `become_user` module key-value.

The boolean variable default values are defined the in the `roles/common/defaults/main.yml` file as the most appropriate location. They can then be overridden when calling the centralized `readiness_checks.yml` task file.

Centralizing the `readiness_checks.yml` task file as a common role allows for task reusability between playbooks, without unnecessary playbook-specific task duplication.

> **IMPORTANT:** Each playbook can be configured to only run the required set of checks, by setting the `vars:` appropriately, for that specific playbook. However, when using the installation shell scripts that run multiple playbooks (serially and independently), the user may notice that the checks are running multiple times as each playbook is run. This is required since each playbook could also be run by itself, outside of the calling shell scripts.

While the tasks are individual `assert` module calls (as looping through the `assert` module would require moving business logic to Ansible variables), the output remains reasonably concise and easy for the user to decipher - due to meaningful task names followed by either the "ok" message or a descriptive `fail_msg`.

Example:

```
...
$ ./install-oracle.sh \
>   --instance-ip-addr ${INSTANCE_IP_ADDR} \
>   --instance-hostname db-server19c \
>   --ora-edition EE \
>   --ora-version 19.3.0.0.0 \
>   --ora-swlib-bucket gs://BUCKET_NAME/19c \
>   --backup-dest "+RECO"
Command used:
./install-oracle.sh --instance-ip-addr 10.2.80.113 --instance-hostname db-server19c --ora-edition EE --ora-version 19.3.0.0.0 --ora-swlib-bucket gs://BUCKET_NAME/19c --backup-dest +RECO

...

Running Ansible playbook: ansible-playbook -i ./inventory_files/inventory_db-server19c_ORCL   check-instance.yml

PLAY [Playbook pre_tasks] **********************************************************************************************************************************************************

TASK [common : Gather target hardware facts] ***************************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Ansible version is 2.9 or higher] *****************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | JMESpath is installed] ****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Hardware architecture is x86_64] ******************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | OS family is RedHat] ******************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | OS version is >= 7.3] *****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Target memory >= 2560] ****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Swap space >= 8192] *******************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Validate privilege escalation ("become_user: root") ability] **************************************************************************************
ok: [db-server19c]
...
```

If a particular condition should fail (which can simulated by changing the minimum memory requirement in the `group_vars/all.yml` file to some artificially high or unreasonable value), a meaningful description message is included in the JSON Ansible output, the tasks fails, and consequently the playbook execution ends:

```
...
PLAY [Playbook pre_tasks] **********************************************************************************************************************************************************

TASK [common : Gather target hardware facts] ***************************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Ansible version is 2.9 or higher] *****************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | JMESpath is installed] ****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Hardware architecture is x86_64] ******************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | OS family is RedHat] ******************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | OS version is >= 7.3] *****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Target memory >= 40000] ***************************************************************************************************************************
fatal: [db-server19c]: FAILED! => {"assertion": "ansible_memtotal_mb >= 40000", "changed": false, "evaluated_to": false, "msg": "Ansible fact detected target memory (MB): 14748"}

PLAY RECAP *************************************************************************************************************************************************************************
db-server19c               : ok=6    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

And if additional readability is required, the user can leverage settings such as `export ANSIBLE_STDOUT_CALLBACK=debug` or `export ANSIBLE_STDOUT_CALLBACK=yaml` resulting in easier to read output such as:

```
...
PLAY [Playbook pre_tasks] **********************************************************************************************************************************************************

TASK [common : Gather target hardware facts] ***************************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Ansible version is 2.9 or higher] *****************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | JMESpath is installed] ****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Hardware architecture is x86_64] ******************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | OS family is RedHat] ******************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | OS version is >= 7.3] *****************************************************************************************************************************
ok: [db-server19c]

TASK [common : readiness_check | Target memory >= 40000] ***************************************************************************************************************************
fatal: [db-server19c]: FAILED! => {
    "assertion": "ansible_memtotal_mb >= 40000",
    "changed": false,
    "evaluated_to": false
}

MSG:

Ansible fact detected target memory (MB): 14748

PLAY RECAP *************************************************************************************************************************************************************************
db-server19c               : ok=6    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

Generally, Ansible tasks should be concise. Showing a meaningful task description along with the `OK:` task status should be sufficient. And consequently, `assert` module `success_msg` usually just provides redundant and cluttering output and is generally not necessary. Consequently, did not include `success_msg` key values in the `assert` module tasks.

---

> **IMPORTANT:** Python version and whether to install Python is currently not included in this initial set of checks. But arguably it could be added to this framework if desired by simply adding it as another assert to the task file.
>
> Currently that is handled in the `check-instance.yml` playbook which is being reviewed in a complementary but separate change.

### Test Prep:

Tests can be run by simulating a failure condition by editing some of the variables used for readiness tests in the `group_vars/all.yml` file.

For example, simulate failure by changing:

1. `minimum_ansible_version` to `3.0`
1. `os_family_supported` to `RedHatX`
1. `os_min_supported_version` to `9.9`
1. `os_supported_architecture` to `aarm64`
1. `os_minimum_memory_mb` to `40000`

Simulated changes can be implemented one at a time or in combination.

Further, test by uninstalling JMESpath (from the Ansible control node):

```bash
sudo yum remove -y python38-jmespath
sudo pip3 uninstall jmespath
```

Separately, test by removing sudo ability from test user (on the target/managed host server). For example, if running in GCE, as root:

```bash
gpasswd --delete USERNAME google-sudoers

# To undo after running toolkit:
#gpasswd --add USERNAME google-sudoers
```

### Test details:

Test with any playbook including the simple, and quick to run `check-instance.yml` playbook:

```bash
export INSTANCE_IP_ADDR=10.2.83.201

./install-oracle.sh \
  --ora-swlib-bucket gs://BUCKET_NAME/19c \
  --check-instance
```

Optionally, experiment with various `ANSIBLE_STDOUT_CALLBACK` levels:

```bash
export ANSIBLE_STDOUT_CALLBACK=default
export ANSIBLE_STDOUT_CALLBACK=debug
export ANSIBLE_STDOUT_CALLBACK=yaml
```

```bash
ansible-playbook -i ./inventory_files/inventory_${INSTANCE_IP_ADDR}_ORCL check-instance.yml -v
```

## Expected Results:

Example output when JMESpath is not installed and with STDOUT callback set to `debug`:

```
...
TASK [common : readiness_check | Ansible version is 3.0 or higher] *****************************************************************************************************************
fatal: [10.2.80.113]: FAILED! => {
    "assertion": "ansible_version.full is version_compare('3.0', '>=')",
    "changed": false,
    "evaluated_to": false
}

MSG:

Ansible 3.0 or higher is required but wasn't detected: install Ansible 3.0 or higher

PLAY RECAP *************************************************************************************************************************************************************************
10.2.80.113                : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

---

Example output when the `ansible_user` does not have sudo ability on the managed host:

```
...
PLAY [Playbook pre_tasks] **********************************************************************************************************************************************************

TASK [common : Gather target hardware facts] ***************************************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | Ansible version is 2.9 or higher] *****************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | JMESpath is installed] ****************************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | OS family is RedHat] ******************************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | OS version is >= 7.3] *****************************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | Target memory >= 2560] ****************************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | Swap space >= 8192] *******************************************************************************************************************************
ok: [10.2.80.113]

TASK [common : readiness_check | Validate privilege escalation ("become_user: root") ability] **************************************************************************************
fatal: [10.2.80.113]: FAILED! => {}

MSG:

Missing sudo password

PLAY RECAP *************************************************************************************************************************************************************************
10.2.80.113                : ok=7    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```


AlexBasinov (approved):

group_vars/all.yml
```diff
@@ -140,3 +140,36 @@ firsttime_connect_user: "customeradmin"
...
 u01_lun: "{{ lookup('env','ORA_U01_LUN')|lower|default('',true) }}"
+
+#  Validations
+minimum_ansible_version: 2.9
+os_family_supported: "RedHat"
+os_min_supported_version: "{% if free_edition %}8{% else %}7.3{% endif %}"
+os_supported_architecture: "x86_64"
+os_minimum_memory_mb: 2560
+
+control_node_checks:
```
mfielding, line 151: Given that these are logic-heavy, can we put these checks into a playbook and include as necessary?

This file is a good place for constants though.

group_vars/all.yml
```diff
@@ -140,3 +140,36 @@ firsttime_connect_user: "customeradmin"
...
+  - check_result: "{{ true if ansible_version.full is version(minimum_ansible_version, '>=') else false }}"
+    check_test: "ansible_version.full is version_compare('{{ minimum_ansible_version }}', '>=')"
+    finding: "Ansible 2.9 or higher is required but wasn't detected: install Ansible 2.9 or higher"
```
mfielding, line 154: Ansible {{ minimum_ansible_version }} or higher...
(In case it's changed in the future)

group_vars/all.yml
mfielding: btw I like how easy these are to read, and concisely written.

group_vars/all.yml
```diff
@@ -140,3 +140,36 @@ firsttime_connect_user: "customeradmin"
...
+  - check_result: "{{ lookup('pipe', ansible_playbook_python ~ ' -c \"import jmespath\" && echo true || echo false') == 'true' }}"
+    check_test: "{{ lookup('pipe', ansible_playbook_python ~ ' -c \"import jmespath\" && echo true || echo false') == 'true' }}"
+    prefix_msg: "jmespath installed "
+    finding: "Wasn't able to run 'import jmespath': install on your Ansible control node"
+
+managed_host_checks:
+  - check_result: "{{ ansible_os_family == os_family_supported }}"
+    check_test: "ansible_os_family == '{{ os_family_supported }}'"
+    finding: "Ansible fact detected OS family: {{ ansible_os_family }}"
```
mfielding, line 163: Will it be clear in the message that the problem is that {{ ansible_os_family }} or not in the list of supported OSes {{ os_family_supported }} ?

mfielding (commented):
Just a quick reminder @simonpane ;  there are a few comments I'd love to see your thoughts on.

group_vars/all.yml
simonpane: Using the Ansible variables is included for this, and all other tasks in the re-worked task-file.  FYI

group_vars/all.yml
simonpane: PR [221](https://github.com/google/oracle-toolkit/pull/221) handles this allowing almost all OSs that are in the right OS family to be used.

group_vars/all.yml
```diff
@@ -146,3 +146,11 @@ firsttime_connect_user: "customeradmin"
 control_node_key_file: "~/.ssh/id_rsa_oracle_toolkit"
 proxy_setup: "{{ lookup('env','ORA_PROXY_SETUP')|lower|default('false',true) }}"
 u01_lun: "{{ lookup('env','ORA_U01_LUN')|lower|default('',true) }}"
+
+#  Environment and infrastructure values (requirements) used for readiness checks
```
mfielding, line 150: Is there a reason we're listing these values here rather than putting in readiness_checks.yml directly?   This is a bit harder to read, but is it related to some sort of override?

brute-cleanup.yml
```diff
@@ -13,14 +13,26 @@
...
+  gather_facts: false
+  pre_tasks:
+    - name: Check environment readiness
+      include_role:
+        name: common
+        tasks_from: readiness_checks.yml
+      vars:
+        control_node_checks: true
+        managed_host_checks: true
+        become_user_check: root
```
mfielding, line 27: Is there a use case in mind for becoming a different user?  If, for example, we wanted to be able to become the {{ grid_user }}, it seems more likely that we'd want readiness_checks.yml to check _both_ root and {{ grid_user }} as separate checks anyway.

brute-cleanup.yml
```diff
@@ -13,14 +13,26 @@
...
+- name: Playbook pre_tasks
+  hosts: dbasm
+  gather_facts: false
+  pre_tasks:
+    - name: Check environment readiness
+      include_role:
+        name: common
+        tasks_from: readiness_checks.yml
+      vars:
+        control_node_checks: true
```
mfielding, line 25: To me, these variables are a bit confusing to read.  A couple alternate proposals I'd like to run by you:

1. Have separate control_node_checks.yml and managed_host_checks.yml and call them in a loop.  And duplicate the become_user check.  (though, separately, I'd prefer _not_ to have root access on the control node if possible)
2. Apply tags like the last example in https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_role_module.html

group_vars/all.yml
simonpane: The reasoning is that these are variables that could possibly change in time (i.e. minimum Ansible version) with a toolkit update, not an user-defined override.

If we define them in the task files then they would be set using the `vars:` key at the task level.  Even though changes would be data/variable specific, it would look like a value modification is to a task file (and task) which should be just business logic.

Alternatively, the task files could be converted into playbooks with a `vars:` section at the top of the playbook.  But these tasks are more suitable as reusable task files with imported tasks as they are called from other (numerous) playbooks.

Another option would be to define them in `common/defaults/main.yml` (or within another role).  But variables in `defaults/main.yml` are supposed to be for configurable options which can be overriden.  Whereas `group_vars/all.yml` is for variables applicable in all circumstances, representing a configuration standard or policy based value.  Rarely, but possibly changed in a future release (i.e. going bak to the minimum Ansible version example).

Anyway, that was the reasoning.  But we can change it to wherever you think it's most appropriate.

brute-cleanup.yml
simonpane: The idea behind the single file implementation is that some tasks may (in the future) be applicable to one or more sets of checks.  Having a single task file both eliminates duplication and confusion.

Kinda hard to come up with a practical example right now so trying to explain through a hypothetical instead.  Say, in the future there were "RAC checks" and "DG checks".  There could be an overlapping "thing" that needs to be checked with an assert task.  So which file would it go into?  Or if the task (check) was added to both there would be duplication.

Whereas if implemented in a single task file, it could easily be handled by the boolean logic of the `when: rac_checks or dg_checks` key-value.

And perhaps the `vars:` section might be less confusing if the variable names were prefixed and changed to `perfrom_control_node_checks` and `perform_managed_host_checks`.

Anyway, just sharing the initial reasoning behind this for your consideration.  If you still feel that having separate files and a loop is better, just let me know and we'll happily adjust it accordingly.

brute-cleanup.yml
simonpane: H'mm that's an interesting thought.   Perhaps the variable should be renamed to indicate that this is really checking for the ability to perform **privilege escalation** to a superuser?  And not generically a user-change as the current variable name might imply.

As for other `become_user:` tasks that are not a privilege escalation to a superuser, they are actually privilege de-escalation with the superuser (root currently) becoming the lower privileged oracle or grid user.

simonpane:
Hi @mfielding , I have merged master into this change to bring it up to date.

When you have a chance can you please review the discussions in the comments above and let me know what's next and what changes you'd like?  Thanks.

brute-cleanup.yml
mfielding: This looks good, thanks!

mfielding (approved):
Apologies for the long time to look at this one.  This looks good.  (And thanks for keeping the branch up to date too!)

PR #216, opened 2025-02-27
feat: Terraform module to deploy GCE instances managed by the toolkit…

GMafra:
## Change Description:
Adding Terraform module to run the toolkit against managed GCE instances.

## Solution Overview:
Terraform module added:

- Creates GCE instances
- local-exec that runs install-oracle.sh with the required parameters
- Example deployment

Gist with full deployment: [link](https://gist.github.com/GMafra/49c1f5d075825a1a95d30b6cf5992372)


asm_disk_config.json
mfielding, line 1: Are we intentionally deleting the example config files?

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+- **IAM Service Account** for managing VM access
+- **SSH Key Management** using Ansible SSH keys for secure access
+- **Custom Metadata Scripts** for VM initialization
+- **Ansible Playbooks** to automate post-deployment configurations
+
+This infrastructure is modular and customizable, allowing you to tailor it to specific application needs or organizational requirements.
+
+---
+
+## Pre-requisites
```
mfielding, line 46: Assuming we're going to be running everything from GCP VMs, is there a "quick start" that automates prerequisite generation?  Might https://cloud.google.com/docs/terraform/install-configure-terraform work?

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+Additionally, you will need:
+
+- A **Google Cloud project** with billing enabled
+- A **Service Account** with appropriate IAM roles for Compute Engine and Storage management
+
+---
+
+## Project Directory Structure
+
+The recommended project directory structure is as follows:
```
mfielding, line 65: I think this is more than the recommended way;  it is the way the git repo is structured.

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+repo-root/
+├── install-oracle.sh               # Main deployment script for Ansible
+├── check-instance.yml
+├── prep-host.yml
+├── install-sw.yml
+├── config-db.yml
+├── config-rac-db.yml
+└── terraform/
+    ├── main.tf                     # Main Terraform configuration
+    ├── ansible-ssh-key             # Private SSH key file
```
mfielding, line 77: Storing security credentials inside git repos seems a bit risky.  What do you think about storing keys outside the repo?  Given that we're in GCP, cloud secret manager is one possibility, or even just ~/.ssh has much better protections.

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+Set your project ID:
+
+```bash
+gcloud config set project PROJECT_ID
+```
+
+2. Generate a SSH key-pair for Ansible
+
+```bash
+ssh-keygen -t ed25519 -C "ansible" -f ansible-ssh-key -N ""
```
mfielding, line 106: This depends on the current directory right?  Maybe save the repo root in a shell variable that van be used for this kind of thing?

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+}
+```
+
+4. Review and Edit Terraform Module Configuration
+
+   Edit `terraform/main.tf` to define your deployment settings. Add however many ASM disks you require in the `asm_disks` section.
+
+   Below is an example configuration:
+
+```terraform
```
mfielding, line 145: This is getting rather large to store inside documentation;  maybe store somewhere in the repo that users can just copy and use directly?

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+#...
+module "oratk_ansible" {
```
mfielding, line 147: Branding-wise, the project is called GCP Toolkit for Oracle, or oracle-toolkit for short.  Not oratk.

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+  source = "./modules/oratk-ansible"
+
+  region                = "REGION"
+  zone                  = "ZONE"
+  project               = "PROJECT_ID"
+  subnetwork            = "SUBNET"
+  service_account_email = "SERVICE_ACCOUNT_EMAIL"
+
+  image_map = {
+    rhel7  = "projects/rhel-cloud/global/images/rhel-7-v20240611"
```
mfielding, line 157: Please don't hard-code image revisions.  https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance suggests names like "debian-cloud/debian-11"

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+module "oratk_ansible" {
+  source = "./modules/oratk-ansible"
+
+  region                = "REGION"
+  zone                  = "ZONE"
+  project               = "PROJECT_ID"
+  subnetwork            = "SUBNET"
+  service_account_email = "SERVICE_ACCOUNT_EMAIL"
+
+  image_map = {
```
mfielding, line 156: Alphabetize?

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+    rhel7  = "projects/rhel-cloud/global/images/rhel-7-v20240611"
+    alma8  = "projects/almalinux-cloud/global/images/almalinux-8-v20241009"
+    rhel8  = "projects/rhel-cloud/global/images/rhel-8-v20241210"
+    rocky8 = "projects/rocky-linux-cloud/global/images/rocky-linux-8-v20250114"
+  }
+
+  instance_name  = "oracle-test"
```
mfielding, line 163: Beyond the example, it would be nice to have more formal documentation of all the possible parameters, descriptions, defaults, warnings etc in the same form as in https://google.github.io/oracle-toolkit/user-guide.html#parameters.  They could even be added to the user guide directly.

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+  instance_count = 1
+  image          = "rhel8"
+  machine_type   = "n2-standard-4"
+  #metadata_startup_script = "gs://BUCKET/SCRIPT.sh"  # Optional - use only if required
+  network_tags   = ["oracle", "ssh"]  # Optional - use only if required
+
+  base_disk_size = 50
+
+  fs_disks = [
```
mfielding, line 172: For someone starting out here, it's difficult to understand what is boilerplate and default, and what needs to be customized.  We could highlight the replacements, but also consider taking out default values (auto_delete, boot, etc) out of the example.

terraform/README.md
mfielding, line 106: Oh, and is there a reason that the terraform (or ansible) can't create these keys by default?

terraform/main.tf
```diff
@@ -0,0 +1,86 @@
...
+  service_account_email = var.service_account_email
+
+  image_map = {
+    rhel7  = "projects/rhel-cloud/global/images/rhel-7-v20240611"
+    alma8  = "projects/almalinux-cloud/global/images/almalinux-8-v20241009"
+    rhel8  = "projects/rhel-cloud/global/images/rhel-8-v20241210"
+    rocky8 = "projects/rocky-linux-cloud/global/images/rocky-linux-8-v20250114"
+  }
+
+  instance_name  = "oracle-test"
```
mfielding, line 31: I'm a bit confused about the difference between example.tfvars and main.tf.  Shouldn't things like the instance name also be in a tfvars faile?

terraform/modules/oratk-ansible/main.tf
```diff
@@ -0,0 +1,122 @@
...
+  subnetwork_project = var.project
+  service_account = {
+    email  = var.service_account_email
+    scopes = ["https://www.googleapis.com/auth/cloud-platform"]
+  }
+
+  machine_type = var.machine_type
+  source_image = lookup(var.image_map, var.image, null)
+  disk_size_gb = var.base_disk_size
+  disk_type    = "pd-balanced"
```
mfielding, line 59: Isn't this parameterized too?

terraform/modules/oratk-ansible/main.tf
```diff
@@ -0,0 +1,122 @@
...
+    metadata_startup_script = var.metadata_startup_script
+    ssh-keys                = "ansible:${file(var.ssh_public_key_path)}"
+  }
+
+  additional_disks = local.additional_disks
+
+  tags = var.network_tags
+}
+
+module "compute_instance" {
```
mfielding, line 73: Do we need firewall rules to access the instance, particularly the Oracle listener?  (not from the public Internet, but at least the local VPC?)

terraform/modules/oratk-ansible/main.tf
```diff
@@ -0,0 +1,122 @@
...
+  subnetwork_project  = var.project
+  num_instances       = var.instance_count
+  hostname            = var.instance_name
+  instance_template   = module.instance_template.self_link
+  deletion_protection = false
+
+  access_config = [
+    {
+      nat_ip       = null
+      network_tier = "STANDARD"
```
mfielding, line 89: PREMIUM is the default network tier;  any reason we're using STANDARD here?

terraform/modules/oratk-ansible/main.tf
```diff
@@ -0,0 +1,122 @@
...
+      user        = "ansible"
+      private_key = file("${var.ssh_private_key_path}")
+      host        = each.value.network_interface[0].access_config[0].nat_ip
+    }
+  }
+
+  provisioner "local-exec" {
+    working_dir = "../"
+    command     = <<-EOT
+      ./install-oracle.sh \
```
mfielding, line 111: bash install-oracle.sh (in case it's not executable)

terraform/README.md
```diff
@@ -0,0 +1,298 @@
...
+- **SSH Key Management** using Ansible SSH keys for secure access
+- **Custom Metadata Scripts** for VM initialization
+- **Ansible Playbooks** to automate post-deployment configurations
+
+This infrastructure is modular and customizable, allowing you to tailor it to specific application needs or organizational requirements.
+
+---
+
+## Pre-requisites
+
```
mfielding, line 47: Maybe an ansible playbook (once ansible itself is installed of course)

terraform/modules/oratk-ansible/main.tf
```diff
@@ -0,0 +1,122 @@
...
+
+    connection {
+      type        = "ssh"
+      user        = "ansible"
+      private_key = file("${var.ssh_private_key_path}")
+      host        = each.value.network_interface[0].access_config[0].nat_ip
+    }
+  }
+
+  provisioner "local-exec" {
```
mfielding, line 108: https://developer.hashicorp.com/terraform/cloud-docs/run/install-software#avoid-installing-extra-software
https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax#provisioners-are-a-last-resort
Is there a better way to provision the toolkit here?

terraform/README.md
gabriel-crosshatch: I think this is outside of the scope of the change, as mentioned on a previous comment, it should be at the user's discretion to set up the box where the toolkit will run from

asm_disk_config.json
gabriel-crosshatch: Not sure I follow, we have examples for all the required files for terraform, we are only deleting state/temp-config related ones, such as terraform.lock.hcl

GMafra: Not sure I follow, we have examples for all the required files for terraform, we are only deleting state/temp-config related ones, such as terraform.lock.hcl

terraform/README.md
GMafra: Updated [4e5c894](https://github.com/google/oracle-toolkit/pull/216/commits/4e5c89434ad8f1083dd52ab1428e6c96de4fbf72)

terraform/README.md
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), ssh key pair is now created, stored on secret manager, added as a local_file resource to be used for install-oracle.sh script and then deleted afterwards.

terraform/README.md
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), ssh key pair is now created, stored on secret manager, added as a local_file resource to be used for install-oracle.sh script and then deleted afterwards.

terraform/README.md
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), renamed

terraform/README.md
GMafra: it was intented to be a map users could costumize, refactored on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a) for simplicity

terraform/README.md
GMafra: I prefer to group/order them by with what it relates to for better readbility

terraform/README.md
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), terraform/main.tf now how a link to this doc, the parameters for install-oracle.sh, for terraform values, defaults and examples, all are present on main.tf as well

terraform/README.md
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), added examples for everything.

terraform/README.md
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), ssh key pair is now created, stored on secret manager, added as a local_file resource to be used for install-oracle.sh script and then deleted afterwards.

terraform/main.tf
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), tfvars is a way to pass variable values for modules, but the way we are intantiating it here doesn't make too much sense to have one, removed, all values to be passed on terraform/main.tf

terraform/modules/oratk-ansible/main.tf
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), added a variable

terraform/modules/oratk-ansible/main.tf
GMafra: if the box running the toolkit is on the same subnet where the Terraform will provision the instances, then no, the default fw rule to allow subnet traffic is enough

terraform/modules/oratk-ansible/main.tf
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a), switched to PREMIUM

terraform/modules/oratk-ansible/main.tf
GMafra: addressed on [7f5e072](https://github.com/google/oracle-toolkit/pull/216/commits/7f5e0723dffde3aa58db18a7f7394810236db25a)

terraform/README.md
GMafra: I think this is outside of the scope of the change, as mentioned on a previous comment, it should be at the user's discretion to set up the box where the toolkit will run from

terraform/modules/oratk-ansible/main.tf
GMafra: Not as of now with Terraform, the [ansible provider](https://github.com/ansible/terraform-provider-ansible) is not mature enough, tried to get it working but just as a general feedback they've been receiving, handling dynamic inventory is not fully supported yet among other limitations

asm_disk_config.json
mfielding: This PR deletes `asm_disk_config.json` and `data_mounts_config.json`;  I'm not sure this was intentional.

GMafra: Oh ok, that is correct, configuration/logic that were on those files is now handled automatically by terraform.

terraform/README.md
mfielding: Let's please create an action item to document (and, at some point, automate) how to satisfy the prerequisites.

terraform/README.md
```diff
@@ -0,0 +1,176 @@
...
+- **Google Cloud SDK** - [Installation Guide](https://cloud.google.com/sdk/docs/install)
+- **Terraform** - [Installation Guide](https://learn.hashicorp.com/tutorials/terraform/install-cli)
+- **Ansible** - [Installation Guide](https://docs.ansible.com/ansible/latest/installation_guide/index.html)
+- **jq** - JSON processor required for handling playbook variables - [Installation Guide](https://stedolan.github.io/jq/download/)
+- **JMESPath** - [Installation Guide](https://pypi.org/project/jmespath/)
+
+Additionally, you will need:
+
+- A **Google Cloud project** with billing enabled
+- A **Service Account** with appropriate IAM roles for Compute Engine and Storage management
```
mfielding, line 59: Can we please list the IAM roles that are required?

terraform/README.md
mfielding: Awesome!

terraform/modules/oratk-ansible/main.tf
mfielding: I assume you're referring to https://cloud.google.com/firewall/docs/firewalls#more_rules_default_vpc ?  These rules include SSH and RDP, but the toolkit creates an Oracle database listener, which is how database clients can reach the database server.  Can we configure a firewall, based on parameters as needed, to allow clients to connect to the configured listener port (under https://google.github.io/oracle-toolkit/user-guide.html#software-installation-parameters)?  Perhaps a list of subnets?

terraform/modules/oratk-ansible/main.tf
mfielding: Ack thanks

asm_disk_config.json
mfielding: I don't think we mean to make Terraform mandatory;  we still want it to work (with defaults) as a command-line tool.

GMafra: files reinstated in [fac654f](https://github.com/google/oracle-toolkit/pull/216/commits/fac654f8699fa8017e26fca72d004337cc872e52)

terraform/README.md
GMafra: added to the backlog. We will need to go through a round of scoping to get the requirements for this task.

terraform/modules/oratk-ansible/main.tf
GMafra: added example FW rule [815e96c](https://github.com/google/oracle-toolkit/pull/216/commits/815e96c677591c8262644490929ddb883eeab622). I don't believe it is necessary to add that functionality on the toolkit module itself, users can source the existing [FW module by google](https://github.com/terraform-google-modules/terraform-google-network/tree/v10.0.0/modules/firewall-rules) to create their rules.

terraform/main.tf
```diff
@@ -0,0 +1,127 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module "oracle_toolkit" {
+  source = "./modules/oracle_toolkit_module"
+  #
+  # Fill in the information below
```
mfielding, line 18: Can these be defined as input variables please (with appropriate descriptions and validation)?  The intent is that users can invoke with a .tfvars or command-line variable settings, should they so choose.  https://developer.hashicorp.com/terraform/language/values/variables

GMafra: I removed tfvars [(see comment)](https://github.com/google/oracle-toolkit/pull/216#discussion_r1985258147) due to user clarity before, do we want the module call at terraform/main.tf to have variables for all the values as well? Followed the same pattern we have for the [sap terraform](https://cloud.google.com/solutions/sap/docs/tf-for-sap) module. We need a final decision on which approach to take here.

terraform/main.tf
```diff
@@ -0,0 +1,127 @@
...
+  # Full list of parameters can be found here https://google.github.io/oracle-toolkit/user-guide.html#parameters
+  # The example below will install Oracle 19c, using the Oracle software stored in a GCS bucket, and will configure the backup destination to be RECO diskgroup.
+  extra_ansible_vars = [
+    "--ora-swlib-bucket BUCKET", # gcs bucket where the Oracle software is stored, example: gs://my-oracle-software/19c 
+    "--ora-version 19",
+    "--backup-dest +RECO"
+  ]
+}
+
+# Firewall rules
```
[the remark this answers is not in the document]
GMafra: [see](https://github.com/google/oracle-toolkit/pull/216#discussion_r1992202227), users should be able to add any sort of composition they want at terraform/main.tf, firewall rulles are not part of the toolkit module itself, so they can add variables/tfvars as they see fit

terraform/main.tf
```diff
@@ -0,0 +1,127 @@
...
+  ##############################################################################
+  ## OPTIONAL SETTINGS
+  ##   - default values will be determined/calculated
+  ##############################################################################
+  # metadata_startup_script = "STARTUP_SCRIPT" # example: gs://BUCKET/SCRIPT.sh
+  # network_tags            = "NETWORK_TAGS"   # example: ["oracle", "ssh"]
+
+  # Full list of parameters can be found here https://google.github.io/oracle-toolkit/user-guide.html#parameters
+  # The example below will install Oracle 19c, using the Oracle software stored in a GCS bucket, and will configure the backup destination to be RECO diskgroup.
+  extra_ansible_vars = [
```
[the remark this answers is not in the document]
GMafra: or just tool_kit_param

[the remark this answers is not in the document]
mfielding: I was actually thinking of putting the commonly-used ansible parameters into first-class variables, so that common installations won't need to use extra_ansible_vars at all.

So we have the `--ora-swlib-bucket`, `--ora-version`, and `--backup-dest` here.  I set `--ora-swlib-path`, but I think the defaults should be fine, as long we consistently label a purpose of "software" in the disk mount.

Also, commonly-used optional parameters.
`--ora-db-name` (validation rule: https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/DB_NAME.html)
`--ora-db-container` (true/false)
`--ntp-pref`, for cloud installs, should be set to 169.254.169.254
`--oracle-rel`, the specific release update (patchlevel) to install, with similar validation as to software versions
`--ora-edition`, either EE, SE, SE2, or FREE
`--ora-listener-port`, the TCP port to listen for client connections
`--ora-redo-log-size`, a number, 100MB, or 100GB

terraform/main.tf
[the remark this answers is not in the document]
mfielding: Yeah, let's shelve this for now, but given that we manage the software firewall for the DB hosts, I don't think managing the cloud firewall for the same hosts would be a big jump.

terraform/main.tf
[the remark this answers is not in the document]
GMafra: addressed on [556c919](https://github.com/google/oracle-toolkit/pull/216/commits/556c9195550fe36619018cd12cc59218aaee5780), ansible params moved into the toolkit module with logic to pass them back to the script if they are set as variables on terraform

terraform/README.md
GMafra: added [f475693](https://github.com/google/oracle-toolkit/pull/216/commits/f475693f5c5298fec26ec1eac580359de31c8983)

mfielding: @GMafra I think this is the last remaining outstanding comment; can we list the actual IAM roles we recommend, without too much over-granting?  From a quick read of https://cloud.google.com/compute/docs/access/iam, would `roles/compute.instanceAdmin` be sufficient?

GMafra: @mfielding Yeah instanceAdmin would be enough, updated [294e626](https://github.com/google/oracle-toolkit/pull/216/commits/294e62698656280316574c7501f1578a1a982bbd)

mfielding: Actually reading the description of the role, I saw:

     If the user will be managing virtual machine instances that are configured to run as a service account, you must also grant the roles/iam.serviceAccountUser role.

I imagine we'll need this too?

GMafra: if the user or CI platform is authenticated with that SA, then it is not needed, `roles/iam.serviceAccountUser` is for when you are trying to assign that specific service account to a service, for example GKE nodes running under a specifc SA for them

terraform/modules/oratk-ansible/main.tf
mfielding: Created backlog item F0021 to track a future refactor that avoids the need for local-exec (and the unstated dependency on a persistent execution location)

PR #221, opened 2025-03-07
Enterprise Linux 8 package installation simplification

simonpane:
Simplify package installation for Enterprise Linux 8 based systems, removing redundant tasks and unused package repo file checks.

Also, minor formatting adjustments of `when:` keys for related tasks to align with the best practice of using multiple lines with multiple "and" conditions.

### DIFF
--- a/roles/ora-host/tasks/main.yml
+++ b/roles/ora-host/tasks/main.yml
@@ -69,49 +69,43 @@
     opts: rw,nosuid,nodev
     state: mounted
 
-- name: Check for existing package manager repo files
-  stat:
-    path: "/etc/yum.repos.d/{{ item }}"
-  register: repo_file_check
-  with_items: "{{ package_repository_files }}"
-  tags: os-packages
-
-- name: Build a list of detected package manager repo files
-  set_fact:
-    repo_files: "{{ repo_file_check.results | selectattr('stat.exists') | map(attribute='item') | list }}"
-  tags: os-packages
-
-- name: List all detected package manager repo files
-  debug:
-    var: repo_files
-    verbosity: 1
-  tags: os-packages
-
-- name: Install Oracle required packages (RHUI/cloud config) for RHEL7
-  package:
-    name: "{{ oracle_required_rpms + oracle_required_rpms_el7 }}"
-    state: present
-    lock_timeout: "{{ pkg_mgr_lock_timeout }}"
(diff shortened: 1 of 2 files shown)


roles/ora-host/tasks/main.yml
```diff
@@ -99,33 +109,18 @@
...
-    - "'redhat.repo' in repo_files or 'rh-cloud.repo' in repo_files"
-  tags: os-packages
-
-- name: Install Oracle required packages for Oracle Linux
+- name: Install Oracle required packages
   package:
     name: "{{ oracle_required_rpms + vars['oracle_required_rpms_el' + ansible_distribution_major_version] }}"
     state: present
     lock_timeout: "{{ pkg_mgr_lock_timeout }}"
-  loop: "{{ repo_files }}"
```
mfielding, line 124: If we now only use repo_files for RHEL7, should we skip the steps to calculate it in non-RHEL7 cases?  Maybe even a block?

mfielding (commented):
I'm a huge fan of simplification... thanks!  Are there any cases we want to test other than the built-in RHEL8 one?

roles/ora-host/tasks/main.yml
simonpane: Yes, that's a very good suggestion @mfielding .  Have implemented as you suggested in commit https://github.com/google/oracle-toolkit/pull/221/commits/18dcc148218338cf63a9e666b7a744ebb0a7de69 - moved all RHEL7 specific steps into a dedicated block.

Re-tested against all major versions with results shown in:
- [oratk-35: RHEL7 Full test run output](https://gist.github.com/simonpane/94962152bf2099428014052d49fc877f)
- [oratk-35: Non-marketplace RedHat 7 Full test run output](https://gist.github.com/simonpane/e7e351a6e2aacbcf8ae8ebecbfad6c4c)
- [oratk-35: RHEL8 Full test run output](https://gist.github.com/simonpane/d512141d417971243fb26c120bec4c53)
- [oratk-35: Oracle Linux 7 Full test run output](https://gist.github.com/simonpane/2ce57437984b619f9049a195fc035ba5)
- [oratk-35: Oracle Linux 8 Full test run output](https://gist.github.com/simonpane/1d7d228645c9b8a9e75cc66e2c849331)

mfielding (approved):
Looks good to me

PR #235, opened 2025-04-22
Switch to ssh-keygen for PEM format and clarify task names

AlexBasinov:
Replace the use of Ansible's openssh_keypair module with a direct call to ssh-keygen in order to fix the following [error](https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/233/oracle-toolkit-install/1914377806712672256):

```
Unsupported parameters for (openssh_keypair) module: backend, private_key_format
Supported parameters include: attributes, backup, comment, content, delimiter, directory_mode, follow, force, group, mode, owner, path, regexp, remote_src, selevel, serole, setype, seuser, size, src, state, type, unsafe_writes
```

The private_key_format and backend parameters are only available in the community.crypto collection (v1.7.0+), which would require users to install an external dependency. Using ssh-keygen directly avoids the need for additional collections.

Also updated task names and variable names to improve clarity.

Testing:
Manually ran presubmit tests on BMS machines. [The modified tasks completed successfully](https://gist.github.com/AlexBasinov/321604e901359fa72a493e48eaa032bd). However, the "rac-gi-install" task failed with the following error:

```
The cluster name (oracle-toolkit-) cannot end with hyphen character (-).
ACTION: Specify a valid cluster name.
```

This appears to be a completely different issue unrelated to the changes in this PR.

### DIFF
--- a/roles/ssh-setup/tasks/main.yml
+++ b/roles/ssh-setup/tasks/main.yml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ---
-- name: Create .ssh dir if already present
+- name: Ensure /home/{{ ssh_user }}/.ssh directory exists with correct permissions
   become: true
   become_user: "{{ ssh_user }}"
   file:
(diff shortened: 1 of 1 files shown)


roles/ssh-setup/tasks/main.yml
```diff
@@ -24,45 +24,47 @@
...
   become_user: "{{ ssh_user }}"
-  openssh_keypair:
-    path: /home/{{ ssh_user }}/.ssh/id_rsa
-    mode: u=rw,go=
-    owner: "{{ ssh_user }}"
-    group: "{{ user_group }}"
-    private_key_format: "pkcs1"
-    backend: "cryptography"
-  register: ssh_key
+  command: ssh-keygen -m PEM -t rsa -b 2048  -f /home/{{ ssh_user }}/.ssh/id_rsa -N ''
```
mfielding, line 30: Can we quote the directory name to avoid too many shenanigans from user-specified inputs?

roles/ssh-setup/tasks/main.yml
```diff
@@ -24,45 +24,47 @@
...
+  args:
+    creates: /home/{{ ssh_user }}/.ssh/id_rsa
+  tags: ssh-keys
+
+- name: Read SSH public key for user '{{ ssh_user }}'
```
mfielding, line 35: In the interest of idiomatic ansible, can we use a lookup plugin for this?  https://docs.ansible.com/ansible/latest/plugins/lookup.html#using-lookup-plugins

mfielding (commented):
Thanks for clarifying variable names etc, and it seems like running ssh-keygen is pretty clean.

Just a few nits hete.

roles/ssh-setup/tasks/main.yml
AlexBasinov: The [lookup plugin](https://docs.ansible.com/ansible/latest/plugins/lookup.html) runs only on the Ansible control machine, not on remote target machines which is what we're doing here.

roles/ssh-setup/tasks/main.yml
AlexBasinov: Done

mfielding (approved):

PR #238, opened 2025-04-25
Optimizations (fixes and enhancements) to RMAN active duplicate tasks

simonpane:
## Change Description:

The change optimizes (added compatibility fixes and other enhancements) the RMAN active-copy phase of the Oracle Data Guard setup tasks.

Specific adjustments include:

1. Defining the standby database name in a new `standby_name` Ansible variable. Previously, the composite of `{{ db_name }}{{ standby_suffix }}` was used. However, it was used in many locations and the number of usages is growing with Data Guard task step improvements. Consequently it makes sense to define, and use the new variable.
1. Update the `active-copy.yml` task file for idempotency. Including first checking the state of an existing database, and only performing the active duplication tasks if an existing, and clusterware registered database is not present. Regardless of whether the setup was performed, or skipped, the state of the standby database is shown at then end of the task file tasks. With these changes, if the database is present and registered with the clusterware, the task-file is now completely idempotent (meaning a PLAY RECAP with `changed=0`).
1. As of Oracle 21c the default location of the password file changes from `$ORACLE_HOME/dbs` to `$ORACLE_BASE/dbs`. Therefore tasks are updated to first check for a password file in a custom (and clusterware registered) location, then the Oracle Home location, and finally the Oracle Base location.
1. Move password file to ASM for standby instance.
1. Perform other database adjustments (from the `db-adjustments` role) to ensure that block change tracking and force logging is enabled on the standby. Otherwise, in the event of a Data Guard switch-over, the newly prompted environment does not match the characteristics of the original database.
1. Listener is only reloaded if changed.
1. Redundant `become:` task keys removed as a clean-up activity. Playbook, and consequently this task-file is run as root and therefore `become` is only required when accompanied by a `become_user` key-value.

## Test Results:

Tested against Oracle Database versions 21c, 19c, 18c, and 12gR2 on Oracle Linux 8:

- [Oracle Database 21c Test Run Output](https://gist.github.com/simonpane/955f406a5a1464cf7f6cdc4c0ccdf9d2)
- [Oracle Database 19c Test Run Output](https://gist.github.com/simonpane/4b7a72fd2acf498375307aa6eada3e0d)
- [Oracle Database 18c Test Run Output](//gist.github.com/simonpane/b82ebf3c2ab58c6084fade0c88c99bdf)
- [Oracle Database 12gR2 Test Run Output](https://gist.github.com/simonpane/d431655554048f360b8e96182ac59631)

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -105,10 +105,9 @@ prereq_option: "{% if oracle_ver_base in ('11.2', '12.1') %}-ignorePrereq{% else
 # section for DR
 
 # db_unique_name identifies uniqueness of a database in Data Guard configuration.
-# By default in the script it set to "_s" so it might be helpful to overwrite the variable
-# and set it to the suffix which identifies location of the standby database
-# (like "_sydney" for example)
-# The same suffix will be "attached" to the db name in DG configuration as well.
+# For the standby database, this is represented by the Ansible "standby_name"
+# variable. This variable can be customized as desired.  For example, using the
+# suffix "_s" or something location specific (i.e. "_sydney").
 standby_suffix: _s
 standby_name: "{{ db_name }}{{ standby_suffix }}"
 
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -110,6 +110,7 @@ prereq_option: "{% if oracle_ver_base in ('11.2', '12.1') %}-ignorePrereq{% else
 # (like "_sydney" for example)
 # The same suffix will be "attached" to the db name in DG configuration as well.
 standby_suffix: _s
+standby_name: "{{ db_name }}{{ standby_suffix }}"
 
 # section size to use during RMAN duplicate
 section_size: 32G
--- a/roles/db-copy/tasks/active-copy.yml
+++ b/roles/db-copy/tasks/active-copy.yml
@@ -38,7 +38,7 @@
         owner: "{{ oracle_user }}"
         group: "{{ oracle_group }}"
 
-    - name: Active-copy | Set random password
+    - name: Active-copy | Generate random password
       set_fact:
         sys_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
 
@@ -82,7 +82,7 @@
       set_fact:
         password_file_name: "{{ password_file if password_file | length > 0 else (oracle_home + '/dbs/orapw' + db_name) if orapw_home_stat.stat.exists else (oracle_base + '/dbs/orapw' + db_name) }}"
 
-    - name: Active-copy | Backup password file from file system
+    - name: Active-copy | Back up password file from file system
       copy:
         src: "{{ password_file_name }}"
         dest: "{{ oracle_home }}/dbs/orapw{{ db_name }}.{{ lookup('pipe', 'date +%Y-%m-%d-%H-%M') }}"
@@ -97,7 +97,7 @@
         - password_file is not search('^\\+')
         - (orapw_home_stat.stat.exists or orapw_base_stat.stat.exists)
 
-    - name: Active-copy | Backup password file from ASM
+    - name: Active-copy | Back up password file from ASM
       shell: |
         set -o pipefail
         asmcmd cp {{ password_file }} {{ grid_home }}/dbs/orapw{{ db_name }}.{{ lookup('pipe', 'date +%Y-%m-%d-%H-%M') }}
@@ -146,7 +146,7 @@
       become_user: "{{ grid_user }}"
       when: standby_listener_update.changed
 
-    - name: Active-copy | Create directories
+    - name: Active-copy | Create audit dump directory
       file:
         path: "{{ oracle_base }}/admin/{{ oracle_sid }}/adump"
         state: directory
@@ -192,7 +192,7 @@
       become_user: "{{ oracle_user }}"
       no_log: true
 
-    - name: Active-copy | Get duplicate script
+    - name: Active-copy | Generate duplicate script
       template:
         src: duplicate.cmd.j2
         dest: "{{ oracle_home }}/dbs/duplicate.cmd"
--- a/roles/db-copy/tasks/active-copy.yml
+++ b/roles/db-copy/tasks/active-copy.yml
@@ -13,212 +13,272 @@
 # limitations under the License.
 
 ---
-- name: Active-copy | Check pmon process
+- name: Active-copy | Check if standby database is already running
   shell: |
-    set -o pipefail
-    ps -ef | ( grep pmon || true ) | ( grep -i {{ oracle_sid }}$ || true ) | ( grep -v grep || true ) | wc -l
+    srvctl status database -d {{ standby_name }} || true
+  environment:
+    ORACLE_HOME: "{{ oracle_home }}"
+    ORACLE_SID: "{{ oracle_sid }}"
+    PATH: "{{ oracle_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+  become: true
+  become_user: "{{ oracle_user }}"
   changed_when: false
-  register: pmon_proc
+  register: standby_status
   tags: active-duplicate
 
-- name: Active-copy | Add oratab entry
-  lineinfile:
-    path: /etc/oratab
-    regexp: '^{{ oracle_sid }}\:'
-    line: "{{ oracle_sid }}:{{ oracle_home }}:N"
-    owner: "{{ oracle_user }}"
-    group: "{{ oracle_group }}"
-  become: true
+- name: Active-copy | Run RMAN duplicate tasks
+  when: "'is running' not in standby_status.stdout"
   tags: active-duplicate
+  block:
+    - name: Active-copy | Add oratab entry
+      lineinfile:
+        path: /etc/oratab
+        regexp: '^{{ oracle_sid }}\:'
+        line: "{{ oracle_sid }}:{{ oracle_home }}:N"
+        owner: "{{ oracle_user }}"
+        group: "{{ oracle_group }}"
 
-- name: Active-copy | Set random password
-  set_fact:
-    sys_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
-  tags: active-duplicate
+    - name: Active-copy | Set random password
+      set_fact:
+        sys_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
 
-- name: Active-copy | Get primary db password file information
+    - name: Active-copy | Get primary db password file information
+      shell: |
+        set -o pipefail
+        srvctl config db -d {{ db_name }} | grep "^Password file"
+      environment:
+        ORACLE_HOME: "{{ oracle_home }}"
+        ORACLE_SID: "{{ oracle_sid }}"
+        PATH: "{{ oracle_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+      delegate_to: primary1
+      become: true
+      become_user: "{{ oracle_user }}"
+      changed_when: false
+      register: srvctl_output
+
+    - name: Active-copy | Password file variable
+      set_fact:
+        password_file: "{{ srvctl_output.stdout | regex_replace('^Password file:') | regex_replace('\\s') }}"
+
+    - name: Active-copy | Search for password file in ORACLE_HOME
+      stat:
+        path: "{{ oracle_home }}/dbs/orapw{{ db_name }}"
+      delegate_to: primary1
+      become: true
+      become_user: "{{ oracle_user }}"
+      register: orapw_home_stat
+      when: password_file | length == 0
+
+    - name: Active-copy | Search for password file in ORACLE_BASE
+      stat:
+        path: "{{ oracle_base }}/dbs/orapw{{ db_name }}"
+      delegate_to: primary1
+      become: true
+      become_user: "{{ oracle_user }}"
+      register: orapw_base_stat
+      when: password_file | length == 0 and not orapw_home_stat.stat.exists
+
+    - name: Active-copy | Set password file name variable
+      set_fact:
+        password_file_name: "{{ password_file if password_file | length > 0 else (oracle_home + '/dbs/orapw' + db_name) if orapw_home_stat.stat.exists else (oracle_base + '/dbs/orapw' + db_name) }}"
+
+    - name: Active-copy | Backup password file from file system
+      copy:
+        src: "{{ password_file_name }}"
+        dest: "{{ oracle_home }}/dbs/orapw{{ db_name }}.{{ lookup('pipe', 'date +%Y-%m-%d-%H-%M') }}"
+        owner: "{{ oracle_user }}"
+        group: "{{ oracle_group }}"
+        mode: "u=wr,go="
+        remote_src: true
+      delegate_to: primary1
+      become: true
+      become_user: "{{ oracle_user }}"
+      when:
+        - password_file is not search('^\\+')
+        - (orapw_home_stat.stat.exists or orapw_base_stat.stat.exists)
+
+    - name: Active-copy | Backup password file from ASM
+      shell: |
+        set -o pipefail
+        asmcmd cp {{ password_file }} {{ grid_home }}/dbs/orapw{{ db_name }}.{{ lookup('pipe', 'date +%Y-%m-%d-%H-%M') }}
+      environment:
+        ORACLE_HOME: "{{ grid_home }}"
+        ORACLE_SID: "{{ asm_sid }}"
+        PATH: "{{ grid_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+      delegate_to: primary1
+      become: true
+      become_user: "{{ grid_user }}"
+      when: password_file is search('^\\+')
+
+    - name: Active-copy | Set sys password for primary db
+      command:
+        argv:
+          - "{{ oracle_home }}/bin/orapwd"
+          - "file={{ password_file_name }}"
+          - "force=y"
+          - "password={{ sys_pass }}"
+      environment:
+        ORACLE_HOME: "{{ oracle_home }}"
+        ORACLE_SID: "{{ oracle_sid }}"
+        PATH: "{{ oracle_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+      delegate_to: primary1
+      become: true
+      become_user: "{{ oracle_user }}"
+      no_log: true
+
+    - name: Active-copy | Add static listener entry
+      lineinfile:
+        path: "{{ grid_home }}/network/admin/listener.ora"
+        regexp: "^SID_LIST_{{ listener_name }}"
+        line: "SID_LIST_{{ listener_name }}=(SID_LIST=(SID_DESC=(GLOBAL_DBNAME={{ standby_name }}{% if db_domain | default('', true) | length > 0 %}.{{ db_domain }}{% endif %})(ORACLE_HOME={{ oracle_home }})(SID_NAME={{ oracle_sid }})))"
+        owner: "{{ grid_user }}"
+        group: "{{ oracle_group }}"
+      register: standby_listener_update
+      become: true
+      become_user: "{{ grid_user }}"
+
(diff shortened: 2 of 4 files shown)


roles/common/defaults/main.yml
```diff
@@ -110,6 +110,7 @@ prereq_option: "{% if oracle_ver_base in ('11.2', '12.1') %}-ignorePrereq{% else
 # (like "_sydney" for example)
 # The same suffix will be "attached" to the db name in DG configuration as well.
 standby_suffix: _s
```
mfielding, line 112: If standby_suffix isn't otherwise used, how about just setting standby_name "{{ db_name }}_s"

It brings up a bigger question though:  if there's a role switch, now _s isn't the standby anymore.  It's the primary.  Not sure if there's an easy way to generalize here;  Orafcle's examples typically use geographic locations, but standbys aren't always in a different place.

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
-- name: Active-copy | Set random password
-  set_fact:
-    sys_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
-  tags: active-duplicate
-
-- name: Active-copy | Get primary db password file information
-  shell: |
-    set -o pipefail
-    srvctl config db -d {{ db_name }} | grep "^Password file"
+    srvctl status database -d {{ standby_name }} || true
```
mfielding, line 18: We plan to add support for filesystem configurations fairly soon, including data guard;  will these changes make it more difficult?  

[oracle@oracle-linux-94 ~]$ srvctl status database -d test
****Unable to retrieve Oracle Clusterware home.
Start Oracle Clusterware stack and try again.

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
-- name: Active-copy | Create auxiliary init file
-  template:
-    src: initaux.ora.j2
-    dest: "{{ oracle_home }}/dbs/init{{ oracle_sid }}.ora"
-    owner: "{{ oracle_user }}"
-    group: "{{ oracle_group }}"
-    mode: "u=wr,go="
-  become: true
-  tags: active-duplicate
+    - name: Active-copy | Backup password file from file system
```
mfielding, line 85: Nit: "Back up" in the verb tense

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
-    host sleep 60
-    EOF
-  environment:
-    ORACLE_HOME: "{{ oracle_home }}"
-    ORACLE_SID: "{{ oracle_sid }}"
-    PATH: "{{ oracle_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
-  become: true
-  become_user: "{{ oracle_user }}"
-  tags: active-duplicate
+    - name: Active-copy | Backup password file from ASM
```
mfielding, line 100: "Back up"

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
-    owner: "{{ oracle_user }}"
-    group: "{{ oracle_group }}"
-    mode: "u=wr,go="
-    remote_src: true
-  delegate_to: primary1
-  become: true
-  become_user: "{{ oracle_user }}"
-  when: password_file is not search('^\\+')
-  tags: active-duplicate
+    - name: Active-copy | Set random password
```
mfielding, line 41: Can we put this step closer to when we actually use sys_pass as input to orapwd?

mfielding: Also, I'd call this "Generate random password";  the setting happens later.

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
+      set_fact:
+        sys_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
```
mfielding, line 43: Do we need a no_log to keep the password out of logs here?

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
+    - name: Active-copy | Reload listener
+      shell: "{{ grid_home }}/bin/lsnrctl reload {{ listener_name }}"
+      environment:
+        ORACLE_HOME: "{{ grid_home }}"
+        PATH: "{{ grid_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+      become: true
+      become_user: "{{ grid_user }}"
+      when: standby_listener_update.changed
+
+    - name: Active-copy | Create directories
```
mfielding, line 149: Specifically, "Create audit dump directory"

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
+        group: "{{ oracle_group }}"
+        mode: "u=wr,go="
+
+    - name: Active-copy | Start auxiliary instance
+      shell: |
+        set -o pipefail
+        {{ oracle_home }}/bin/sqlplus -s -L / as sysdba <<EOF
+        startup nomount pfile={{ oracle_home }}/dbs/init{{ oracle_sid }}.ora force
+        alter system register;
+        host sleep 60
```
mfielding, line 171: Do we really need to sleep here?  I though alter system register was an immediate operation.

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
+      become_user: "{{ oracle_user }}"
 
-- name: Active-copy | Add Oracle Restart configuration
+    - name: Active-copy | Set sys password for auxiliary instance
+      command:
+        argv:
+          - "{{ oracle_home }}/bin/orapwd"
+          - "file={{ password_file_name }}"
+          - "force=y"
+          - "password={{ sys_pass }}"
```
mfielding, line 186: Even with no_log, command lines get into a lot of logs, notably /var/log/secure because of sudo.  If it possible to feed in the password via stdin?

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
+      environment:
+        ORACLE_HOME: "{{ oracle_home }}"
+        ORACLE_SID: "{{ oracle_sid }}"
+        PATH: "{{ oracle_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+      become: true
+      become_user: "{{ oracle_user }}"
+      no_log: true
+
+    - name: Active-copy | Get duplicate script
```
mfielding, line 195: I'd call this "Generate duplicate script"

roles/db-copy/tasks/active-copy.yml
```diff
@@ -13,212 +13,272 @@
...
+        group: "{{ oracle_group }}"
+        mode: "u=wr,go="
+      become: true
+      become_user: "{{ oracle_user }}"
+
+    - name: Active-copy | Duplicate primary database
+      command:
+        argv:
+          - "{{ oracle_home }}/bin/rman"
+          - 'target "sys/{{ sys_pass }}@//{{ lookup(''env'', ''PRIMARY_IP_ADDR'') }}:{{ listener_port | default(1521, true) }}/{{ db_name }}{% if db_domain | default('''', true) | length > 0 %}.{{ db_domain }}{% endif %}"'
```
mfielding, line 209: I think this command would leak passwords right?  Will we get enough troubleshooting info if we no-log this one?  Any other ways to protect the password:  wallet?

roles/db-copy/templates/initaux.ora.j2
```diff
@@ -1,2 +1,2 @@
-db_name={{ db_name }}
-db_unique_name={{ db_name }}{{ standby_suffix }}
+*.db_name={{ db_name }}
```
mfielding, line 1: Can/should we quote here?

simonpane: I don't think that either parameter here needs to be quoted? Parameter documentation for reference:
- [DB_NAME](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/DB_NAME.html)
- [DB_UNIQUE_NAME](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/DB_UNIQUE_NAME.html)

In both cases the special characters allowed are limited to `_`, `$`, and `#` and double-quotes (at least are removed).  I can try with single quotes but I don't recall that ever been necessary - have you run into such use cases before @mfielding ?

roles/db-copy/tasks/active-copy.yml
simonpane: Changed in commit https://github.com/google/oracle-toolkit/pull/238/commits/0f893870603fdd54533889e9aca770544c39426b.

roles/db-copy/tasks/active-copy.yml
simonpane: As an initial change, renamed task as suggested in commit https://github.com/google/oracle-toolkit/pull/238/commits/0f893870603fdd54533889e9aca770544c39426b.  Will separately review moving task or the task's action.

roles/db-copy/tasks/active-copy.yml
simonpane: Changed as suggested in commit https://github.com/google/oracle-toolkit/pull/238/commits/0f893870603fdd54533889e9aca770544c39426b.

But we're not really generating, unless you would count it as "generating" from a template with jinja2 substitutions - we could use "Copy" as the verb if you prefer, or leave it with the updated value from that commit - whichever you like @mfielding ?

roles/db-copy/tasks/active-copy.yml
simonpane: Added the `no_log: true` key-value to start with in commit https://github.com/google/oracle-toolkit/pull/238/commits/9883ff7b693c0ca1487ee67843c005be94c6a00c.  And RMAN automatically excludes passwords from the RMAN output log.

Will explore using the Secure External Password Store wallet as an alternative.  That solution may have other pros and cons however.  For example, the wallet itself needs a password.

roles/db-copy/tasks/active-copy.yml
simonpane: Re-ordered the tasks in commit https://github.com/google/oracle-toolkit/pull/238/commits/9883ff7b693c0ca1487ee67843c005be94c6a00c to collocate all tasks involving the sys password.  And to process them later in the task-file (so that they don't run needlessly if the task file was to fail on an earlier task).

roles/db-copy/tasks/active-copy.yml
simonpane: Agreed.  Not sure why that was originally added to the toolkit.  Removed and re-tested in commit https://github.com/google/oracle-toolkit/pull/238/commits/9883ff7b693c0ca1487ee67843c005be94c6a00c.

roles/db-copy/tasks/active-copy.yml
simonpane: Added `no_log: true` in commit https://github.com/google/oracle-toolkit/pull/238/commits/9883ff7b693c0ca1487ee67843c005be94c6a00c.

Yes we can alternatively use the shell module and redirect stdin with the password.  However doing so could risk exposing the stdin redirection in the output of commands such as `ps`.  So I'm not sure I'm understanding the benefit of the stdin redirection alternative - perhaps you can shed some additional light on the benefit @mfielding ?

Another option is to generate the password and place the password in a temporary file all on the managed host.  Then re-direct the contents on that file.  And of course remove the tempfile afterwards.  That would prevent the password from appearing in the command log file or stdout.  The risk would now be that the password appears in a plain text file on the managed host and that file could possibly not be removed should the task or playbook fail for some reasons.

What are your thoughts @mfielding ?

roles/db-copy/tasks/active-copy.yml
simonpane: Actually in testing, it may not be immediate - it may take the listener up to a minute to show the new service.  Continuing to test.

simonpane: Issue likely due to incorrectly specified `LOCAL_LISTENER` parameter.  Meaning that the listener cannot be found by the register command and hence the delay for 60 seconds behind required.

mfielding: Thanks for the detailed investigation here;  definitely interesting that the listener won't even bind to the network port until a minute after reload command.  This is pretty deep into the Oracle design so unlikely to have an easy fix.

Would you mind making the delay a bit more explicit by adding a "pause" playbook step rather than a sleep in the shell?

roles/db-copy/tasks/active-copy.yml
mfielding: I guess we're already relying on srvctl anyway (just adding it later), so we might as well leverage it.  One follow-up question though:  If we do a "|| true" and presumably continue unconditionally, do we get a benefit from checking the status at all here?

roles/db-copy/tasks/active-copy.yml
mfielding: The goal here would not to expose the password to `/var/log/secure` and any downstream logging systems that might consume it?  

The first question is:  is this a problem at all:  does `no_log` have a way of hiding the password from sudo logs in /var/log/secure already? 

Some hacky unix-y ideas though:
- Write the plaintext pwd file to a tmpfs location like /run or /dev/shm, and attempt to clean up on failure (rescue?)
- Use a non-stdin file descriptor, say &9
- Use a buffered, named pipe (https://unix.stackexchange.com/questions/439196/change-buffer-size-of-named-pipe has some ideas) that can only be read once

I don't know much about Ansible Vault, but might be overly complex for what we want?  That being said, in the roadmap, we'll likely want to make the randomly-generated password available somehow, and that mgiht be a cool way to do it.

roles/db-copy/tasks/active-copy.yml
mfielding: We're applying the template: doing some type of computation to result in what I'd call a generated configuration file.  That being said, "Create duplicate script" is fine;  Gemini suggests "templating a file" or "render template"

roles/db-copy/tasks/active-copy.yml
mfielding: This should be fine for now.  If users invoke --debug, then logfiles are likely to contain passwords.  I've created https://github.com/google/oracle-toolkit/pull/239 to note this in our docs.

roles/db-copy/templates/initaux.ora.j2
mfielding: Works for me, thanks!

roles/db-copy/tasks/active-copy.yml
simonpane: I'm fine to leave it as "Generated", just just brainstorming with you but now that you've explained your reasoning, it makes sense. Will leave it as "Generate duplicate script".

roles/db-copy/tasks/active-copy.yml
simonpane: Hi @mfielding .

Testing around your initial question of:

> The first question is: is this a problem at all: does no_log have a way of hiding the password from sudo logs in /var/log/secure already?

**TL;DR : Ansible is already protecting the sensitive information from logs - but it is exposed through ANSIBLE_DEBUG.**

--- 
## Details through a small test (using Ansible 2.9):

Created the following playbook:
```yaml
$ cat test-pwd.yml
---
- name: Test Ansible password handling
  hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - name: Test without the no_log key
      shell: |
        echo "Hello world"
      become: true
      become_user: root
    - name: Test including the no_log key
      shell: |
        echo "Hello foobar"
      become: true
      become_user: root
      no_log: true
```

The goal is to ensure that the word `foobar` is not exposed.

### Results:

The word `foobar` is not exposed to STDOUT/STDERR, even when verbosity is turned up to `-vvvv`.  I can post the output of that but am excluding it as it's long (obviously).

Lines added to `/var/log/secure`:

```plaintext
May  1 12:56:26 pa-o-78 sudo[1848]:    pane : TTY=pts/1 ; PWD=/home/pane ; USER=root ; COMMAND=/bin/sh -c 'echo BECOME-SUCCESS-jebjkztnejabbguzncfokdhaixdigqif ; /usr/bin/python3.8 /home/pane/.ansible/tmp/ansible-tmp-1746118586.6286778-1836-278138127080721/AnsiballZ_command.py'
May  1 12:56:26 pa-o-78 sudo[1848]: pam_unix(sudo:session): session opened for user root by pane(uid=1000)
May  1 12:56:27 pa-o-78 sudo[1848]: pam_unix(sudo:session): session closed for user root
May  1 12:56:27 pa-o-78 sudo[1868]:    pane : TTY=pts/1 ; PWD=/home/pane ; USER=root ; COMMAND=/bin/sh -c 'echo BECOME-SUCCESS-ukndfvsjkvlwjmyupgpljqdnvovwjzhi ; /usr/bin/python3.8 /home/pane/.ansible/tmp/ansible-tmp-1746118587.0591378-1856-182393557877976/AnsiballZ_command.py'
May  1 12:56:27 pa-o-78 sudo[1868]: pam_unix(sudo:session): session opened for user root by pane(uid=1000)
May  1 12:56:27 pa-o-78 sudo[1868]: pam_unix(sudo:session): session closed for user root
```

But command and arguments are typically shown in `/var/log/messages` (testing on EL8):

```
May  1 12:56:27 pa-o-78 ansible-command[1851]: Invoked with _raw_params=echo "Hello world"#012 _uses_shell=True warn=True stdin_add_newline=True strip_empty_ends=True argv=None chdir=None executable=None creates=None removes=None stdin=None
```

As we can see, our task commands were not exposed at all to `/var/log/secure`.

The first task was echoed to `/var/log/message` and hence we do see the string "**Hello World**".  But nothing was echoed for the second tasks where the `no_log: true` key was used.

Double checking:
```bash
# pwd
/var/log
# grep -ri foobar
#
```

However, as per the Ansible documentation, the sensitive information is still exposed when using [ANSIBLE_DEBUG](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#envvar-ANSIBLE_DEBUG).

Simple illustration that confirms the documentation:
```bash
$ export ANSIBLE_DEBUG=true

$ ansible-playbook test-pwd.yml -vvvv 1> test-pwd.out
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not
match 'all'

$ grep -i foobar test-pwd.out
{"cmd": "echo \"Hello foobar\"\n", "stdout": "Hello foobar", "stderr": "", "rc": 0, "start": "2025-05-01 13:07:58.112305", "end": "2025-05-01 13:07:58.116653", "delta": "0:00:00.004348", "changed": true, "invocation": {"module_args": {"_raw_params": "echo \"Hello foobar\"\n", "_uses_shell": true, "warn": true, "stdin_add_newline": true, "strip_empty_ends": true, "argv": null, "chdir": null, "executable": null, "creates": null, "removes": null, "stdin": null}}}
  2149 1746119278.13433: done with _execute_module (command, {'_raw_params': 'echo "Hello foobar"\n', '_uses_shell': True, 'warn': True, '_ansible_check_mode': False, '_ansible_no_log': True, '_ansible_debug': True, '_ansible_diff': False, '_ansible_verbosity': 4, '_ansible_version': '2.9.27', '_ansible_module_name': 'command', '_ansible_syslog_facility': 'LOG_USER', '_ansible_selinux_special_fs': ['fuse', 'nfs', 'vboxsf', 'ramfs', '9p', 'vfat'], '_ansible_string_conversion_action': 'warn', '_ansible_socket': None, '_ansible_shell_executable': '/bin/sh', '_ansible_keep_remote_files': False, '_ansible_tmpdir': '/home/pane/.ansible/tmp/ansible-tmp-1746119277.887138-2149-243474737033880/', '_ansible_remote_tmp': '~/.ansible/tmp'})
```

Ansible recommendation from [How do I keep secret data in my playbook?](https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#how-do-i-keep-secret-data-in-my-playbook) is to use [Ansible Vault](https://docs.ansible.com/ansible/latest/vault_guide/index.html) as you suggested.

---

This all being said, we will have additional requirements to use a password for `dgmgrl` with more recent versions of Oracle (21c and greater) where we must connect using a password. 

The Ansible Vault solution could be considered.  And if pursued, we would have to consider the integration for Terraform deployments.

mfielding (approved):

PR #252, opened 2025-06-05
oracle-toolkit: clean up TFA faster and more forcefully

sahsagar-google:
Development:
1. Replaced the slower tfactl uninstall command with a more robust, forceful cleanup logic.

2. Added a best-effort task to stop and disable the init.tfa and oracle-tfa.service units.

3. Updated pkill tasks to also terminate processes running from the /opt/oracle.ahf/ directory.

4. Updated file removal tasks to also delete the /opt/oracle.ahf directory and its associated systemd service file.

Testing:
1. Setup: Installed and started AHF/TFA on a target Oracle Linux 8 node to establish a running state.

2. Execution: Ran only the modified TFA cleanup tasks using an Ansible tag to avoid impacting the database installation.

3. Verification: Confirmed that all TFA processes were terminated and all AHF-related files and service definitions were successfully deleted.

4. Reboot Test: Confirmed a clean system reboot with no TFA-related errors in the system journal.

Bug: b/397424987

### DIFF
--- a/roles/brute-ora-cleanup/tasks/main.yml
+++ b/roles/brute-ora-cleanup/tasks/main.yml
@@ -174,20 +174,25 @@
     - cluster_name is not defined
     - not free_edition
 
-- name: Uninstall TFA service
-  command: "{{ grid_home }}/bin/tfactl uninstall -local -silent"
-  environment:
-    PATH: "{{ grid_home }}/perl/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
-    PERL5LIB: "{{ grid_home }}/perl/lib"
-  register: uninstall_tfa
+- name: (Best Effort) Ensure TFA services are stopped and disabled
+  ansible.builtin.service:
+    name: "{{ item }}"
+    state: stopped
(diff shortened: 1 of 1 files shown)


roles/brute-ora-cleanup/tasks/main.yml
```diff
@@ -271,9 +277,11 @@
     - "/etc/sysconfig/oracleasm.rpmsave"
     - "/etc/sysconfig/oracledrivers.conf"
     - "/etc/systemd/system/oracle-ohasd.service"
+    - "/etc/systemd/system/oracle-tfa.service"
```
mfielding, line 280: This change doesn't affect the in-memory systemd config.  But if we already stopped the service, maybe it doesn't matter?

mfielding (approved):

roles/brute-ora-cleanup/tasks/main.yml
sahsagar-google: Yep, you are correct since we already stopped the service and deleted all its files, that leftover config is harmless and just disappears on reboot.

PR #256, opened 2025-06-06
Enable OS Login at the instance level

AlexBasinov:
Enable OS Login at the instance level to avoid requiring project-wide changes.

### DIFF
--- a/docs/terraform.md
+++ b/docs/terraform.md
@@ -87,10 +87,6 @@ Upload the ZIP file to your GCS bucket:
 ```bash
 gsutil cp /tmp/oracle-toolkit.zip gs://your-bucket-name/
 ```
-
-### 5. OS Login is enabled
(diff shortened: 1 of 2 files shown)


mfielding (changes requested):
I think we need to also enable OS login on the control node;  otherwise the startup script there will fail.

mfielding (commented):
Changes look good, but in my own tests, they didn't actually work:  the instance got created without enable-oslogin metadata, and thus didn't have OS login enabled;   can you test with a project that has OS login disabled, and see if an instance like this actually gets it enabled?

mfielding:
> Changes look good, but in my own tests, they didn't actually work: the instance got created without enable-oslogin metadata, and thus didn't have OS login enabled; can you test with a project that has OS login disabled, and see if an instance like this actually gets it enabled?

Aaand I fixed my tests;  it _is_ working for me now:  https://gist.github.com/mfielding/d69c6b7b8d94d9c645bec3648894bcab

mfielding (approved):

PR #258, opened 2025-06-07
Add roles/logging.logWriter to service account permissions

nasmart:
Permission to write to Cloud Logging is required.


mfielding (approved):
Thanks for catching this!

PR #265, opened 2025-06-12
Install Google Cloud's Agent for Compute Workloads

AlexBasinov:
This PR adds support for installing the Google Cloud's Agent for Compute Workloads on the database VM.

New flags introduced:

--install_workload_agent: controls agent installation (default: false).
--oracle_metrics_secret: fully qualified Secret Manager secret containing the database user's password, used by the agent for metric collection.

A new google_cloud_cli.yml playbook installs the gcloud CLI (RHEL-only for now) to retrieve the secret containing database user password used by the agent, and workload_agent.yml installs the agent itself (also RHEL-only). The password is fetched early during the ora-host role for validation, then again post-DB install to create the monitoring user and grant permissions. Agent configuration will come in a separate PR to keep this one focused and easier to review.

[Ansible execution log](https://gist.github.com/AlexBasinov/55e73575caa3c93651589dd9db866dfe)

### DIFF
--- a/config-db.yml
+++ b/config-db.yml
@@ -45,10 +45,24 @@
         - lookup('env', 'PRIMARY_IP_ADDR') is not defined or lookup('env', 'PRIMARY_IP_ADDR') | length == 0
       tags: primary-db
 
+    - name: Skip intallation of Google Cloud Agent for Compute Workloads if non-RHEL OS detected
+      debug:
+        msg: "Non-RHEL OS detected. Skipping the installation of Google Cloud Agent for Compute Workloads."
+      when: ansible_os_family != "RedHat"
+      tags: workload-agent
+
+    - name: Skip intallation of Google Cloud Agent for Compute Workloads if non-GCE environment detected
+      debug:
+        msg: "Non-GCE environment detected. Skipping the installation of Google Cloud Agent for Compute Workloads."
+      when: ansible_facts['product_name'] != "Google Compute Engine"
+      tags: workload-agent
+
     - include_role:
         name: workload-agent
       when:
         - install_workload_agent | bool
+        - ansible_facts['product_name'] == "Google Compute Engine"
+        - ansible_os_family == "RedHat"
       tags: workload-agent
 
     - include_role:
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -113,7 +113,7 @@ arch_bu_start_min: "{{ lookup('env','ARCHIVE_BACKUP_MIN')|default('30',true) }}"
 
 # workload-agent related variables
 install_workload_agent: "{{ lookup('env','INSTALL_WORKLOAD_AGENT')|lower|default(false) }}"
-oracle_metrics_secret: "{{ lookup('env','ORACLE_METRICS_SECRET') }}"
+oracle_metrics_secret: "{{ lookup('env','ORACLE_METRICS_SECRET')|lower|default('', true)}}"
 
 
 ## End of customizable Global Variables
--- a/install-oracle.sh
+++ b/install-oracle.sh
@@ -254,9 +254,6 @@ SWAP_BLK_DEVICE_PARAM=".*"
 COMPATIBLE_RDBMS="${COMPATIBLE_RDBMS:-0}"
 COMPATIBLE_RDBMS_PARAM="^[0-9][0-9]\.[0-9].*"
 
-INSTALL_WORKLOAD_AGENT="${INSTALL_WORKLOAD_AGENT:-FALSE}"
-INSTALL_WORKLOAD_AGENT_PARAM="^(TRUE|FALSE)$"
-
 ORACLE_METRICS_SECRET="${ORACLE_METRICS_SECRET}"
 ORACLE_METRICS_SECRET_PARAM="^projects/[^/]+/secrets/[^/]+/versions/[^/]+$"
 
@@ -274,7 +271,7 @@ GETOPT_OPTIONAL="$GETOPT_OPTIONAL,backup-start-hour:,backup-start-min:,archive-b
 GETOPT_OPTIONAL="$GETOPT_OPTIONAL,ora-swlib-type:,ora-swlib-path:,ora-swlib-credentials:,instance-ip-addr:,primary-ip-addr:,instance-ssh-user:"
 GETOPT_OPTIONAL="$GETOPT_OPTIONAL,instance-ssh-key:,instance-hostname:,ntp-pref:,inventory-file:,compatible-rdbms:,instance-ssh-extra-args:"
 GETOPT_OPTIONAL="$GETOPT_OPTIONAL,help,validate,check-instance,prep-host,install-sw,config-db,debug,allow-install-on-vm,skip-database-config,swap-blk-device:"
-GETOPT_OPTIONAL="$GETOPT_OPTIONAL,install-workload-agent:,oracle-metrics-secret:"
+GETOPT_OPTIONAL="$GETOPT_OPTIONAL,install-workload-agent,oracle-metrics-secret:"
 GETOPT_LONG="$GETOPT_MANDATORY,$GETOPT_OPTIONAL"
 GETOPT_SHORT="h"
 
@@ -527,8 +524,7 @@ while true; do
     shift
     ;;
   --install-workload-agent)
-    INSTALL_WORKLOAD_AGENT="$2"
-    shift
+    INSTALL_WORKLOAD_AGENT=true
     ;;
   --oracle-metrics-secret)
     ORACLE_METRICS_SECRET="$2"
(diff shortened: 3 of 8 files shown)


terraform/terraform.tfvars.example
```diff
@@ -22,6 +22,8 @@ project_id                   = "your-project-id"
 subnetwork                   = "default"
 vm_service_account           = "your-vm-service-account@your-project-id.iam.gserviceaccount.com"
 control_node_service_account = "your-control-node-service-account@your-project-id.iam.gserviceaccount.com"
+db_password_secret           = "projects/your-project/secrets/your-secret/versions/latest"
```
mfielding, line 25: oracle_metrics_secret right?  Do we want an example for install_workload_agent while we're at it?

roles/ora-host/tasks/main.yml
```diff
@@ -117,6 +117,35 @@
...
+    file: google_cloud_cli.yml
+  when:
+    - cluster_type != "RAC"
+    - install_workload_agent | bool
+    - oracle_metrics_secret | length > 0
+  tags: workload-agent
+
+- name: Fetch and validate workload-agent's database user password
+  command: gcloud --quiet secrets versions access {{ oracle_metrics_secret }}
+  register: wlm_agent_password
```
mfielding, line 139: Do we use this variable for anything?  It seems that we re-fetch the password when we actually create the user.  (this might be fine, as we do want to fail fast)

roles/workload-agent/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+
+---
+- name: Fetch workload agent user password from Secret Manager
+  command: gcloud --quiet secrets versions access {{ oracle_metrics_secret }}
+  register: wlm_agent_password
+  changed_when: false
+  no_log: true
+  tags: workload-agent
+
+- name: Create Oracle user for Google Cloud Workload Agent
```
mfielding, line 23: Agent for Compute Workloads, technically

group_vars/all.yml
```diff
@@ -111,6 +111,11 @@ arch_bu_start_min: "{{ lookup('env','ARCHIVE_BACKUP_MIN')|default('30',true) }}"
 # backup_mount_src: "192.168.185.82:/backups" #an nfs share of some sort
 # backup_mount_path: "/backups" #an nfs mount point
 
+# workload-agent related variables
+install_workload_agent: "{{ lookup('env','INSTALL_WORKLOAD_AGENT')|lower|default(swlib_path_default,false) }}"
```
mfielding, line 115: What's swlib_path_default here?  Isn't this a boolean?

roles/ora-host/tasks/main.yml
```diff
@@ -117,6 +117,35 @@
     - not (ansible_distribution == 'RedHat' and ansible_distribution_major_version | int == 7)
   tags: os-packages
 
+- name: Install Google Cloud Workload Agent
+  include_tasks:
+    file: workload_agent.yml
+  when:
+    - cluster_type != "RAC"
```
mfielding, line 124: Why don't we install the workload agent on RAC hosts?  Instead, might we want to check if we're running on Google Cloud?

roles/workload-agent/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+  become: true
+  become_user: "{{ oracle_user }}"
+  shell: |
+    set -o pipefail
+    sqlplus -s -L / as sysdba <<EOF
+    WHENEVER SQLERROR EXIT 1;
```
mfielding, line 29: Assuming Ansible propagates the exit code, can we do a WHENEVER SQLERROR EXIT SQL.SQLCODE?  Given that we have no_log, there's very little detail about any errors that may have happened :-)

docs/user-guide.md
```diff
@@ -2223,6 +2223,22 @@ instance is created.</td>
 <td>Run with the Ansible debugging flag enabled.  With debugging
 enabled, passwords appear in logfiles.</td>
 </tr>
+<tr>
+<td></td>
+<td><p><pre>
+--install_workload_agent
+</pre></p></td>
+<td></td>
+<td>Whether to install <a href="https://github.com/GoogleCloudPlatform/workloadagent">Google Cloud Agent for Compute Workloads</a></td>
```
mfielding, line 2232: Better link: https://cloud.google.com/compute/docs/instances/agent-for-compute-workloads

group_vars/all.yml
AlexBasinov: this was a copy-paste mistake

roles/workload-agent/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    CREATE USER {{ workload_agent_username }} IDENTIFIED BY {{ wlm_agent_password.stdout }};
```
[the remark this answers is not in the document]
AlexBasinov: I've added a new Ansible task to validate the fetched password against a regex that restricts it to a safe set of characters.

roles/ora-host/tasks/main.yml
AlexBasinov: Done.

terraform/terraform.tfvars.example
AlexBasinov: Accidentally included a variable from another PR while working on both. Fixed now.

roles/ora-host/tasks/main.yml
```diff
@@ -117,6 +117,55 @@
     - not (ansible_distribution == 'RedHat' and ansible_distribution_major_version | int == 7)
   tags: os-packages
 
+- name: Check if running on Google Cloud
```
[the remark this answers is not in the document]
AlexBasinov: I've replaced this task with the following condition:

ansible_facts['product_name'] == "Google Compute Engine"

group_vars/all.yml
```diff
@@ -111,6 +111,11 @@ arch_bu_start_min: "{{ lookup('env','ARCHIVE_BACKUP_MIN')|default('30',true) }}"
 # backup_mount_src: "192.168.185.82:/backups" #an nfs share of some sort
 # backup_mount_path: "/backups" #an nfs mount point
 
+# workload-agent related variables
+install_workload_agent: "{{ lookup('env','INSTALL_WORKLOAD_AGENT')|lower|default('false',true) }}"
```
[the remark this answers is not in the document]
AlexBasinov: Fixed.

config-db.yml
```diff
@@ -45,6 +45,15 @@
         - lookup('env', 'PRIMARY_IP_ADDR') is not defined or lookup('env', 'PRIMARY_IP_ADDR') | length == 0
       tags: primary-db
 
+    - include_role:
+        name: workload-agent
+      when:
+        - ansible_facts['product_name'] == "Google Compute Engine"
+        - create_db | bool
+        - install_workload_agent | bool
+        - oracle_metrics_secret | length > 0
```
mfielding, line 54: We want to install the agent (without metrics) even if oracle_metrics_secret isn't supplied

config-db.yml
```diff
@@ -45,6 +45,15 @@
         - lookup('env', 'PRIMARY_IP_ADDR') is not defined or lookup('env', 'PRIMARY_IP_ADDR') | length == 0
       tags: primary-db
 
+    - include_role:
+        name: workload-agent
+      when:
+        - ansible_facts['product_name'] == "Google Compute Engine"
```
mfielding, line 51: If not running on Google Cloud, this step will be silently skipped;  maybe worth a warning or message of some type?

config-db.yml
```diff
@@ -45,6 +45,15 @@
...
+        - create_db | bool
```
mfielding, line 52: We can install the agent even if not creating a database in the toolkit;  it can detect databases migrated in the future.  We can't set up metric collection though, naturally.

AlexBasinov:
FYI I moved the agent installation playbook from roles/ora-host/workload_agent.yml to roles/workload-agent/install.yml so that all agent-related tasks (installation, password retrieval, and configuration) are grouped in a single role for better organization and simplicity

config-db.yml
AlexBasinov: Done

config-db.yml
AlexBasinov: Done. PTAL ?

group_vars/all.yml
```diff
@@ -111,6 +111,11 @@ arch_bu_start_min: "{{ lookup('env','ARCHIVE_BACKUP_MIN')|default('30',true) }}"
 # backup_mount_src: "192.168.185.82:/backups" #an nfs share of some sort
 # backup_mount_path: "/backups" #an nfs mount point
 
+# workload-agent related variables
+install_workload_agent: "{{ lookup('env','INSTALL_WORKLOAD_AGENT')|lower|default(false) }}"
+oracle_metrics_secret: "{{ lookup('env','ORACLE_METRICS_SECRET') }}"
```
mfielding, line 116: We've generally defaulted optional parameters to '' to avoid undefined variable errors.  Not sure if it's strictly necessary though.

install-oracle.sh
```diff
@@ -254,6 +254,13 @@ SWAP_BLK_DEVICE_PARAM=".*"
 COMPATIBLE_RDBMS="${COMPATIBLE_RDBMS:-0}"
 COMPATIBLE_RDBMS_PARAM="^[0-9][0-9]\.[0-9].*"
 
+INSTALL_WORKLOAD_AGENT="${INSTALL_WORKLOAD_AGENT:-FALSE}"
```
mfielding, line 257: Rather than accepting a string value, defaulting, and validating, it might be cleaner to handle as a boolean, as we do for --no-patch.  It also allows a more fluent command-line.

terraform/variables.tf
```diff
@@ -232,6 +232,22 @@ variable "gcs_source" {
   }
 }
 
+variable "install_workload_agent" {
+  description = "Whether to install workload-agent on the database VM."
+  type        = bool
+  default     = true
```
mfielding, line 238: We still have the parameter default discrepancy: true in terraform, false in the toolkit.  Is this intentional?

terraform/variables.tf
```diff
@@ -232,6 +232,22 @@ variable "gcs_source" {
...
+}
+
+variable "oracle_metrics_secret" {
+  description = "Fully qualified name of the Secret Manager secret that stores the Oracle database user's password. This user is specifically configured for the workload-agent to enable metric collection."
+  type        = string
```
mfielding, line 243: Unless we require this parameter to be specified, we need a default, even if it's null or an empty string (and also validation that will succeed in this case)

terraform/variables.tf
AlexBasinov: FIxed

mfielding (approved):

PR #267, opened 2025-06-13
Refactor Terraform storage

mfielding:
Yes, yet another refactor, and yet another breaking change.  With this change, we become more prescriptive about disk layout, enforcing a 5-disk layout:
* Boot disk, of course
* Swap disk
* Oracle binaries (/u01)
* Oracle data (currently ASM)
* Oracle fast recovery area (currently ASM)

So the old `asm_disks` and `fs_disks` will no longer be exposed as variables.

We also remove disk names, since they can now be statically created. Sizes are more consistent:  all numbers, and having a unit (usually `_gb`) in the name

Sample ouput (via serial port capture of a Terraform run):  https://gist.github.com/mfielding/21fc66b73b3e9ee824850399e015ca8a


terraform/variables.tf
```diff
@@ -159,18 +147,54 @@ variable "ora_release" {
...
+  default     = 50
+}
+
+variable "swap_disk_type" {
+  description = "The type of the swap disk for the database VM."
   type        = string
   default     = "hyperdisk-balanced"
 }
 
+variable "binaries_disk" {
```
AlexBasinov, line 174: 'binary_disk' sounds a bit confusing, at least to me. it's not immediately clear what binaries it's referring to.
Would you consider renaming it to something clearer like 'ora_home_disk' or 'ora_base_disk' or just 'ora_disk' or maybe 'u01_disk'?

AlexBasinov: ora_bin_disk?

AlexBasinov: oradata_disk?

mfielding: Updated to `oracle_home_disk`

AlexBasinov (approved):

PR #271, opened 2025-06-16
Delete public SSH key from the OS login profile

AlexBasinov:
Remove public SSH key from the control node's service account OS Login profile.

Reason:
If many deployments are created (e.g., ~50), we risk hitting the [32KiB OS Login key limit](https://cloud.google.com/compute/docs/troubleshooting/troubleshoot-os-login#invalid_argument), which would cause the startup script to fail.

[Log from startup script](https://gist.github.com/AlexBasinov/7584dda5ff04ce2b7a9a1db25fb7d375)


mfielding (approved):
Nice catch!

PR #279, opened 2025-06-24
Add support for multi-node Data Guard deployments

AlexBasinov:
Adds support for deploying multi-node Oracle Data Guard configurations on GCE using Terraform.

For multi-instance setups, users provide zone1, zone2, subnetwork1 and subnetwork2. region, zone and network input variables have been removed from variables.tf.

Replaced the use of the terraform-google-modules/vm/google modules with native resources google_compute_instance_template and google_compute_instance_from_template to avoid hardcoded suffixes like -001/-002. Single-instance VMs use var.instance_name directly, and multi-node VMs get -1 and -2 suffixes.

Continue using instance templates to retain automatic disk name generation and support for the existing local.additional_disks structure.

Common flags used by oracle-install.sh are now constructed in Terraform and passed to startup.sh to keep them consistent across both single- and multi-instance deployments.

[log from multi-instance deployment](https://gist.github.com/AlexBasinov/f7671a5795d4b4dcb1f600a17eb42658)


AlexBasinov:
The most recent log from the startup script for the two-node Data Guard deployment: https://gist.github.com/AlexBasinov/88ad361702086f2622d376a21aca7512

terraform/main.tf
```diff
@@ -79,70 +79,149 @@ locals {
...
-  source_image_family  = var.source_image_family
-  source_image_project = var.source_image_project
-  disk_size_gb         = var.boot_disk_size_gb
-  disk_type            = var.boot_disk_type
-  auto_delete          = true
+  dynamic "disk" {
+    for_each = local.additional_disks
+    content {
+      auto_delete       = lookup(disk.value, "auto_delete", null)
+      boot              = lookup(disk.value, "boot", null)
```
mfielding, line 140: If we use `boot_disk_type` and `boot_disk_size_gb` to define the boot disk parameters, perhaps we should remote the option to make an additional disk a boot disk.

terraform/main.tf
```diff
@@ -79,70 +79,149 @@ locals {
...
+      device_name       = lookup(disk.value, "device_name", null)
```
mfielding, line 141: If using multiple nodes, we need to disambiguate the device names for node 1 and node 2, right?

terraform/main.tf
```diff
@@ -79,70 +79,149 @@ locals {
...
+  }
+
 }
 
 resource "random_id" "suffix" {
   byte_length = 4
 }
 
+locals {
+  oracle_nodes = [
```
mfielding, line 185: database_vm_nodes?   I think adding the term "oracle" here might be confusing.

terraform/main.tf
```diff
@@ -79,70 +79,149 @@ locals {
...
+}
+
 resource "google_compute_instance" "control_node" {
   project      = var.project_id
   name         = "${var.control_node_name_prefix}-${random_id.suffix.hex}"
   machine_type = var.control_node_machine_type
-  zone         = var.zone
-
+  zone         = var.zone1
+
```
mfielding, line 224: Trailing spaces

terraform/scripts/setup.sh.tpl
```diff
@@ -104,29 +97,51 @@ EOF
...
+
+  if [[ "$num_nodes" -eq 1 || "$node_name" == *"-1" ]]; then
+    echo "Configuring PRIMARY or SINGLE instance: $node_name, IP: $node_ip, Zone: $node_zone"
+    bash install-oracle.sh \
+    --cluster-type NONE \
+    --instance-ip-addr "$node_ip" \
+    --instance-ssh-user "$ssh_user" \
+    --instance-ssh-key /root/.ssh/google_compute_engine \
+    ${common_flags}
+  elif [[ "$node_name" == *"-2" ]]; then
```
mfielding, line 133: How to we make sure that this installation happens after the primary has been successfully created?

roles/workload-agent/tasks/main.yml
```diff
@@ -18,9 +18,22 @@
...
   tags: workload-agent
 
-- name: Configure Google Cloud Agent for Compute Workloads to collect Oracle metrics and send them to Google Cloud Monitoring
+- name: Create a database user for the Google Cloud Agent for Compute Workloads to collect Oracle metrics
   include_tasks:
-    file: metric_collection.yml
+    file: create_db_user.yml
   when:
     - oracle_metrics_secret | length > 0
+    - "'primary1' in group_names"
```
mfielding, line 26: This is the only mention of `primary1` I can find;  where to we set this?

terraform/main.tf
AlexBasinov: Not necessarily. GCE automatically assigns names to disks that are auto-created from an instance template using the following pattern:

INSTANCE_NAME + "-" + DEVICE_NUMBER


Here's an example:


```
$ gcloud compute instances describe data-guard-1 --zone=us-central1-b --format="value(disks[].source.basename())"
data-guard-1;data-guard-1-1;data-guard-1-2;data-guard-1-3;data-guard-1-4

$ gcloud compute instances describe data-guard-2 --zone=us-central1-c --format="value(disks[].source.basename())"
data-guard-2;data-guard-2-1;data-guard-2-2;data-guard-2-3;data-guard-2-4
```

I realize this numbering might not be very descriptive. We could probably define the disk names in the template based on their function, using something like:


INSTANCE_NAME + "-" + local.additional_disks.device_name

Let me look into how to set that up.

roles/workload-agent/tasks/main.yml
AlexBasinov: There are a few places where the "primary1" host is used to run tasks only on the primary node. For example: https://github.com/search?q=repo%3Agoogle%2Foracle-toolkit%20primary1&type=code

From what I understand, when install-oracle.sh is run a second time with --primary-ip <ip_address> to set up the standby node, it should generate a new inventory file and overwrite the existing one. This inventory defines two groups:

https://github.com/google/oracle-toolkit/blob/7e6e6efd47a67cf66d9c5b5fd90d18ce2b964050/install-oracle.sh#L1001

```
[${INSTANCE_HOSTGROUP_NAME}]
${INSTANCE_HOSTNAME} ansible_ssh_host=${INSTANCE_IP_ADDR} ${COMMON_OPTIONS}

[primary]
primary1 ansible_ssh_host=${PRIMARY_IP_ADDR} ${COMMON_OPTIONS}
```


The [primary] group (with the primary1 host) is then available in the inventory. This lets us target tasks conditionally. The idea was to run the "Create a database user for the Agent" task only on the node in the "primary" group, using this when clause:

- "'primary1' in group_names" 

However, looking more closely at the [last test run](https://gist.github.com/AlexBasinov/88ad361702086f2622d376a21aca7512), I see that this task never actually ran. I'll dig into it further to debug what's going on.

terraform/main.tf
```diff
@@ -79,70 +79,149 @@ locals {
...
+      subnetwork = var.subnetwork1
+    }
+    node2 = {
+      name       = "${var.instance_name}-2"
+      zone       = var.zone2
+      subnetwork = var.subnetwork2
+    }
+  } : {
+    default = {
+      name       = var.instance_name
```
mfielding, line 100: Just had a thought here:  in the future, might we run into a situation where someone wants to add a standby to an existing single-instance database?  Would we change the hostname of the existing primary at that point?  Or just leave it at ${var.instance_name} and a new ${var.instance_name}-2?  The alternative would be to name even single-node as ${var.instance_name}-1.

terraform/main.tf
```diff
@@ -79,70 +79,161 @@ locals {
...
+}
+
+resource "google_compute_instance_template" "default" {
+  name         = "${var.instance_name}-${local.template_suffix}"
+  project      = var.project_id
+  machine_type = var.machine_type
+
+  network_interface {
+    # gets overridden during instance creation
+    network = "default"
```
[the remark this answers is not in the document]
AlexBasinov: In this case "terraform apply" will fail.
I updated this to always use var.subnetwork1 in the instance template. 

PTAL?

terraform/main.tf
AlexBasinov: I renamed this variable to "database_vm_nodes"

terraform/scripts/setup.sh.tpl
AlexBasinov: I refactored startup.sh to run in two passes: first it sets up the primary node, and if there's more than one node, it then pings the primary to ensure it's reachable before setting up any standby nodes. We also no longer rely on the order of items in the database_vm_nodes list. instead, we use the "role" parameter to determine each node's role. 
PTAL?

terraform/main.tf
AlexBasinov: Sorry I marked this as resolved too soon. I just wanted to add some visual separation between the scheduling {} block and the basic parameter settings.

AlexBasinov:
Addressed comments. Test run is in progress. Will share startup script log here when the test run is complete.

AlexBasinov:
The most recent log from the startup script: https://gist.github.com/AlexBasinov/f7671a5795d4b4dcb1f600a17eb42658

PR #281, opened 2025-06-25
Validate actual `vm.nr_hugepages` values

mfielding:
## The problem

While testing memory allocations for https://github.com/google/oracle-toolkit/pull/266, I realized that, if we
attempt to allocate more hugepages than the system can provide, the
command actually succeeds, allocating as many hugepages as it can.  This
is not a good outcome for us:  not only will the DB creation eventually
fail, but the install can hang due to swapping, since all free memory
has been allocated to hugepages.

## The solution

This change runs the `sysctl` OS command to get the current value, and
asserts that it matches what we asked for.

## Sample output
https://gist.github.com/mfielding/8fc8c85a1bd62240e33c72341ee16748

### DIFF
--- a/roles/ora-host/tasks/hugepages.yml
+++ b/roles/ora-host/tasks/hugepages.yml
@@ -38,7 +38,7 @@
 - name: hugepages | Assert that vm.nr_hugepages matches the expected value of {{ v_hugepages_required }}
   assert:
     that:
-      - v_actual_hugepages_allocated.stdout | int == v_hugepages_required
+      - v_actual_hugepages_allocated.stdout | int == v_hugepages_required | int
     fail_msg: "Requested {{ v_hugepages_required }} hugepages to accommodate ora_sga_target_mb={{ ora_sga_target_mb }}, but only got {{ v_actual_hugepages_allocated.stdout }}.  Your requested SGA memory size is likely too large for the available memory."
 
 - name: hugepages | Capture transparent hugepage status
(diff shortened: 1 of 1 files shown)


roles/ora-host/tasks/hugepages.yml
```diff
@@ -18,14 +18,29 @@
...
     state: present
     sysctl_set: true
     reload: true
 
+- name: hugepages | Fetch actual value of vm.nr_hugepages
+  command: sysctl -n vm.nr_hugepages
+  register: v_actual_hugepages_allocated
+  changed_when: false
+
+- name: hugepages | Assert that vm.nr_hugepages matches the expected value of {{ v_hugepages_required }}
```
AlexBasinov, line 38: So from what I see, we first set vm.nr_hugepages, and only then verify that the actual value matches the expected one.

I think it might be more cleaner to check available memory (grep MemAvailable /proc/meminfo) before attempting to set vm.nr_hugepages and failing early. Although It's a minor detail and doesn't impact functionality, so I'm not blocking this. LGTM

AlexBasinov (approved):

roles/ora-host/tasks/hugepages.yml
mfielding: Fair point.  Now that we have a prerequisite check playbook, I'll create a PR to try and add an earlier validation too.

PR #293, opened 2025-07-23
Add double quotes to ensure the query doesn't fail if the password contains special characters

AlexBasinov:
Add double quotes around {{ result.stdout }} to make sure the sql query doesn't fail if the password contains special characters (@, +, ~, *).

Rename the "fail" tasks to indicate which specific password validation failed.

Add a separate "fail" task in create_db_user.yml, similar to the one in validate_passwords.yml, for consistency.

Test results:

[Successful run](https://gist.github.com/AlexBasinov/8b2f40a2e85d265fc709b54b67c3c0d7)

### DIFF
--- a/roles/ora-host/tasks/validate_passwords.yml
+++ b/roles/ora-host/tasks/validate_passwords.yml
@@ -16,8 +16,8 @@
 - name: Fetch workload-agent's user password
   command: gcloud --quiet secrets versions access {{ oracle_metrics_secret }}
   register: agent_pass_result
-  changed_when: false # the command's exit code is checked in a separate task to show a clear error message.
-  failed_when: false 
+  changed_when: false 
+  failed_when: false # the command's exit code is checked in a separate task to show a clear error message.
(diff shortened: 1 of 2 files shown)


mfielding (approved):
Looks good;  we want to make sure we don't have literal double quotes " in the password, and I think the validation will catch those.

The sample output doesn't contain passwords of course, but I'll assume that `projects/abasinov-test/secrets/workload-agent-user-password/versions/latest` does contain some of the characters that were causing issues.

Will you be looking at the RMAN standby creation later?

AlexBasinov:
>Looks good; we want to make sure we don't have literal double quotes " in the password, and I think the validation will catch those.

Correct.

> but I'll assume that `projects/abasinov-test/secrets/workload-agent-user-password/versions/latest` does contain some of the characters that were causing issues.

Yes, that's correct.

>Will you be looking at the RMAN standby creation later?

Yes, I'll address this in a separate PR.

PR #296, opened 2025-07-24
Avoid overwriting SYS user password on primary node in Data Guard setup on GCE

AlexBasinov:
The PR updates the standby database creation process to use the user-provided SYS password from Secret Manager when running on GCE instead of generating a new one and overwriting the primary's password.

Key changes:

- If db_password_secret is provided, the SYS password is retrieved from that Secret Manager entry.

- If db_password_secret is not provided, a random password is generated (this behavior remains unchanged).

- Moved the RMAN connection string from an inline command task into a cmdfile to avoid shell interpretation issues caused by special characters in the password.

Internal bug: b/424271292

Test results:

https://gist.github.com/AlexBasinov/982dfe7a38577fd51ce43b46fd404cc5

### DIFF
--- a/roles/db-copy/tasks/active-copy.yml
+++ b/roles/db-copy/tasks/active-copy.yml
@@ -294,6 +294,7 @@
         state: absent
       become: true
       become_user: "{{ oracle_user }}"
+      ignore_errors: true
 
 - name: Active-copy | Capture standby database state
   shell: |


AlexBasinov:
> The current default is to automatically generate passwords; this is particularly useful in environments that do not have access to Google Cloud Secret Manager. Is there a way we can allow these users to still create data guard standby configurations?

I addressed this in the latest commit. PTAL?

AlexBasinov:
Test results from the latest commit: https://gist.github.com/AlexBasinov/982dfe7a38577fd51ce43b46fd404cc5

roles/db-copy/tasks/active-copy.yml
```diff
@@ -225,6 +243,13 @@
       become_user: "{{ oracle_user }}"
       no_log: true
```
mfielding, line 244: Now that we don't have a password on the command line, can we get away without no_log?
The RMAN command does a lot of work, and it's tough to troubleshoot without the output.  (It does create a log, but doesn't do anything with it AFAICT)

roles/db-copy/tasks/active-copy.yml
```diff
@@ -225,6 +243,13 @@
...
 
+    - name: Delete duplicate.cmd file
```
mfielding, line 246: Is there a way to delete this even if the rman command exits with error?

AlexBasinov: moved this task inside "Always:" block to make sure it runs even if previous tasks failed.

roles/db-copy/tasks/active-copy.yml
```diff
@@ -270,6 +287,14 @@
       become: true
       become_user: "{{ oracle_user }}"
 
+  always:
```
mfielding, line 290: It's also possible that we had a failure before the file was created;  consider either a block for just the template creation and active copy, or an ignore_errors on the deletion.

PR #297, opened 2025-07-24
parameterize control node deletion

AlexBasinov:
Add a delete_control_node Terraform input variable to control whether the control node deletes itself after deployment. By default, the node is removed, but setting the flag to false preserves it for debugging.

Test results:

delete_control_node was set to false: https://gist.github.com/AlexBasinov/ffe96cb192237bc24b9bfafacb9c1959

### DIFF
--- a/terraform/scripts/setup.sh.tpl
+++ b/terraform/scripts/setup.sh.tpl
@@ -17,8 +17,10 @@ cleanup() {
   if ! gcloud --quiet compute os-login ssh-keys remove --key-file=/root/.ssh/google_compute_engine.pub; then
     echo "WARNING: Failed to remove SSH key. Continuing with instance deletion."
   fi
-  echo "Deleting '$control_node_name' GCE instance in zone '$control_node_zone' in project '$control_node_project_id'..."
-  gcloud --quiet compute instances delete "$control_node_name" --zone="$control_node_zone" --project="$control_node_project_id"
+  if [[ ${delete_control_node} == "true" ]]; then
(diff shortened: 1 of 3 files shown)


terraform/scripts/setup.sh.tpl
```diff
@@ -17,8 +17,10 @@ cleanup() {
   if ! gcloud --quiet compute os-login ssh-keys remove --key-file=/root/.ssh/google_compute_engine.pub; then
     echo "WARNING: Failed to remove SSH key. Continuing with instance deletion."
   fi
-  echo "Deleting '$control_node_name' GCE instance in zone '$control_node_zone' in project '$control_node_project_id'..."
-  gcloud --quiet compute instances delete "$control_node_name" --zone="$control_node_zone" --project="$control_node_project_id"
+  if [[ ${delete_control_node} == "true" ]]; then
```
mfielding, line 20: Can we double-quote this "${delete_control_node}" ?  I don't know if there are any other ways to make it obvious that this is a template variable, not a shell variable.

mfielding: What do you think about evaluating it at the template level ie 

  % if delete_control_node %
  echo "Deleting '$control_node_name' GCE instance in zone '$control_node_zone' in project '$control_node_project_id'..."
  gcloud --quiet compute instances delete "$control_node_name" --zone="$control_node_zone" --project="$control_node_project_id"
  % endif %

AlexBasinov: Good idea!

mfielding (approved):

PR #300, opened 2025-08-01
Send a signal to Ansible_logs on startup failure and update Ansible logging plugin

AlexBasinov:
Startup script changes:

  - A JSON log entry is now sent to the `Ansible_logs` Cloud Logging if the startup script fails.

  - The script now uses the modern `gcloud storage cp` instead of `gsutil cp`.

  - All curl commands now use the -fsS flags:
      -f ensures the script fails on HTTP errors.
      -s suppresses the progress meter.
      -S enables error messages when -s is used.
  
  - The script now uses the following idiom for command execution to ensure it exits on failure:
  
  ```
  response="$(command)" || {
    echo "error message"
    exit 1
  }
  ```
  This prevents the script from silently continuing when a command fails.

- apt-get commands now include `--quiet` and `--assume-yes` flags to reduce log noise.

- use `unzip -qq` to suppresses the output of filenames as they are extracted to reduce log noise.

- use `kill "$heartbeat_pid"` to allow the gcloud process terminate gracefully and handle its own cleanup. Also suppress any output or errors.

- Set DEBIAN_FRONTEND=noninteractive to prevent warnings like `dpkg-preconfigure: unable to re-open stdin: No such file or directory` which occur when stdin isn't available in non-interactive environments.

- Added SIGINT and SIGTERM to trap so cleanup runs on interrupt or termination, not just on EXIT.

Ansible logging plugin changes:

  - A new `error_message` field has been added to the jsonPayload. It is populated with either `result.stderr` or `result.msg` depending on the module that fails.
  
  - Set the `state` field in the jsonPayload to either "failed" or "success" based on the status of each Ansible task.

Changes in the "ora-host" role:

- Install Google Cloud cli when either `oracle-metrics-secret` or `db-password-secret` flags are provided.
- Update "Fail if gcloud command failed" Ansible tasks to log the exit code.

### DIFF
--- a/roles/ora-host/tasks/validate_passwords.yml
+++ b/roles/ora-host/tasks/validate_passwords.yml
@@ -26,7 +26,7 @@
 
 - name: Fail if gcloud command failed
   fail:
-    msg: "Command failed: stderr={{ agent_pass_result.stderr }}"
+    msg: "Command failed with rc={{ agent_pass_result.rc }}: stderr={{ agent_pass_result.stderr }}"
   when: 
     # Skip this task if the fetch task was skipped.
     # agent_pass_result.skipped will be true if the fetch task didn't run due to 'when:' conditions.
@@ -56,7 +56,7 @@
 
 - name: Fail if gcloud command failed
   fail:
-    msg: "Command failed: stderr={{ sys_pass_result.stderr }}"
+    msg: "Command failed with rc={{ sys_pass_result.rc }}: stderr={{ sys_pass_result.stderr }}"
   when:
     # Skip this task if the fetch task was skipped.
     # sys_pass_result.skipped will be true if the fetch task didn't run due to 'when:' conditions.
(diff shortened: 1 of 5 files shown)


roles/ora-host/tasks/validate_passwords.yml
```diff
@@ -56,7 +56,7 @@
 
 - name: Fail if gcloud command failed
   fail:
-    msg: "Command failed: stderr={{ sys_pass_result.stderr }}"
+    msg: "Command failed with rc={{ sys_pass_result.rc }}: stderr={{ sys_pass_result.stderr }}"
```
mfielding, line 59: Can you make the same update to https://github.com/google/oracle-toolkit/blob/44605a7786c359d41bb236a717cdec6458844899/roles/workload-agent/tasks/create_db_user.yml#L26 too?

terraform/scripts/setup.sh.tpl
```diff
@@ -1,32 +1,50 @@
 #!/bin/bash
 
-control_node_name="$(curl -s http://metadata.google.internal/computeMetadata/v1/instance/name -H 'Metadata-Flavor: Google')"
```
mfielding, line 3: Any thoughts about a simple "set -e" to bail out on any error, just to make things a bit shorter?

mfielding (approved):

mfielding (approved):

PR #301, opened 2025-08-05
Replace single-instance with Data Guard test and refactor test runner script

AlexBasinov:
The PR replaces the single-instance presubmit test with the Data Guard test, since Data Guard is a superset of the single-instance test and exercises the same code paths in the oracle-toolkit.

Key changes:

- Renamed `single-instance-on-gcp.sh` to `data-guard-on-gcp.sh` to reflect that it now deploys a Data Guard setup. 

- Renamed a `single-instance.tfvars` to `data-guard.tfvars` , and added the `zone2` and `subnetwork2` inputs to enable a two-node setup.

- Removed logging of raw Terraform-created resources as their names are not easily traceable to actual cloud resources.

- Improved log streaming by running `gcloud logging tail` in a loop to workaround its 1-hour session limit. This is necessary for longer-running Data Guard deployments. Also, the logging loop is run in a separate process group using `setsid` program. This allows us to send SIGTERM to the entire process group using `kill -TERM -- -$tail_pgid_leader` from the cleanup() function. This makes sure that all child background processes (such as gcloud) are properly terminated when the main script exits. If `setsid` is not used, and the loop runs in a regular sub-shell, those child processes can continue running in the background. This prevents the container from shutting down properly, causing the Prow job to hang and eventually reach the 4-hour timeout.

-  curl command now uses the `-fsS` flags:
  -f ensures the script fails on HTTP errors.
  -s suppresses the progress meter.
  -S enables error messages when -s is used.

- The `data-guard-on-gcp.sh` script now uses the following idiom for command execution to ensure it exits on failure:

```
response="$(command)" || {
  echo "error message"
  exit 1
}
```

- Improve log readability by:

  - Formatting timestamps using gcloud's `date()` [transform](https://cloud.google.com/sdk/gcloud/reference/topic/projections) function.

  - Removing "startup-script: " string from log lines using gcloud's `sub()` [transform](https://cloud.google.com/sdk/gcloud/reference/topic/projections) function to reduce noise.

Test Results:

https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/301/oracle-toolkit-install-data-guard-on-gcp/1953125605675372544


mfielding (dismissed):
We have a bit of a catch-22 as the presubmit tests are being renamed in the same PR as some major refactoring of the script. So I'm not sure if this will work, but I guess we can always revert if needed.

mfielding (commented):
Reviewing the data guard execution https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/301/oracle-toolkit-install-data-guard-on-gcp/1952740418227539968 I noticed:

* `ERROR: (gcloud.compute.ssh) Users instance [control-node-sa@gcp-oracle-benchmarks.iam.gserviceaccount.com:importSshPublicKey] is the subject of a conflict: Multiple concurrent mutations were attempted. Please retry the request. - maybe a race condition when we create the primary/standby VMs simultaneously?  I didn't see any failures directly tied to this though.
* `2025-08-05 16:17:10	TASK [db-copy : Active-copy | Set sys password for auxiliary instance] *********
2025-08-05 16:17:10	fatal: [192.168.10.62]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}`  Not sure what happened here, but it's probably worth the same output treatment we recently added to secret manager commands

AlexBasinov:
> Reviewing the data guard execution https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/301/oracle-toolkit-install-data-guard-on-gcp/1952740418227539968 I noticed:
> 
> * `2025-08-05 16:17:10	TASK [db-copy : Active-copy | Set sys password for auxiliary instance] ********* 2025-08-05 16:17:10	fatal: [192.168.10.62]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}`  Not sure what happened here, but it's probably worth the same output treatment we recently added to secret manager commands



The `[db-copy : Active-copy | Set sys password for auxiliary instance]` task failed with the following error:

`OPW-00029: Password complexity failed for SYS user : Password must contain at least 1 special character.`

So the SYS password was missing special characters that are required by the `orapwd` tool.  Note that `validate_passwords.yml` task file doesn't prevent this type of issues because its validation pattern (`^[a-zA-Z0-9@+~*]{1,30}$`) only restricts input to a set of allowed characters but does not enforce the password complexity required by the `orapwd` tool. Should we update the `validate_passwords.yml` to enforce password complexity to prevent similar errors in the future?

AlexBasinov:
> Reviewing the data guard execution https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/301/oracle-toolkit-install-data-guard-on-gcp/1952740418227539968 I noticed:
> 
> * `ERROR: (gcloud.compute.ssh) Users instance [[control-node-sa@gcp-oracle-benchmarks.iam.gserviceaccount.com](mailto:control-node-sa@gcp-oracle-benchmarks.iam.gserviceaccount.com):importSshPublicKey] is the subject of a conflict: Multiple concurrent mutations were attempted. Please retry the request. - maybe a race condition when we create the primary/standby VMs simultaneously?  I didn't see any failures directly tied to this though.


The failed connection attempts appear to be caused by a race condition from multiple concurrent requests originating from multiple concurrent presubmit tests. However, this is not a critical failure because the `gcloud compute ssh` command is run in a loop with automatic retries, so these errors should not cause significant problems.

AlexBasinov:
> > Reviewing the data guard execution https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/301/oracle-toolkit-install-data-guard-on-gcp/1952740418227539968 I noticed:
> > 
> > * `2025-08-05 16:17:10	TASK [db-copy : Active-copy | Set sys password for auxiliary instance] ********* 2025-08-05 16:17:10	fatal: [192.168.10.62]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}`  Not sure what happened here, but it's probably worth the same output treatment we recently added to secret manager commands
> 
> The `[db-copy : Active-copy | Set sys password for auxiliary instance]` task failed with the following error:
> 
> `OPW-00029: Password complexity failed for SYS user : Password must contain at least 1 special character.`
> 
> the SYS password was missing special characters that are required by the `orapwd` tool. Note that `validate_passwords.yml` task file doesn't prevent this type of issues because its validation pattern (`^[a-zA-Z0-9@+~*]{1,30}$`) only restricts input to a set of allowed characters but does not enforce the password complexity required by the `orapwd` tool. Should we update the `validate_passwords.yml` to enforce password complexity to prevent similar errors in the future?

I updated the password to include a special character, and then ran the test manually. This time the test completed successfully: https://gist.github.com/AlexBasinov/9fcd8835b8094cd25547dae82619e9af

AlexBasinov:
@mfielding, FYI, here's what happens to log streaming when gcloud logging tail is restarted

...
2025-08-05 22:33:46     TASK [ora-host : Fetch SYS database password] **********************************
2025-08-05 22:33:46     ok: [192.168.10.194]
The maximum duration for tail has been met. The command may be repeated to continue. (Session has run for the maximum allowed duration of 1h. To continue, start a new session with the same request.)
[2025-08-05 22:33:51] Tail session ended with exit code . Restarting...
[2025-08-05 22:33:51] Starting gcloud logging tail session...
Initializing tail session.
2025-08-05 22:34:02
2025-08-05 22:34:02     TASK [ora-host : Add OS groups] ************************************************
2025-08-05 22:34:02     changed: [192.168.10.194] => (item={'group': 'oinstall', 'gid': 801})
2025-08-05 22:34:02     changed: [192.168.10.194] => (item={'group': 'dba', 'gid': 802})
...

AlexBasinov:
> From https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/301/oracle-toolkit-install-data-guard-on-gcp/1952986897739419648:
> 
> ```
> Control node's startup script completed successfully.
> Stopping gcloud logging tail process 157
> ...
> Deleted deployment [presubmit-dg-1952986897739419648].
> ...
> {"component":"entrypoint","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:169","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.ExecuteProcess","level":"error","message":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2025-08-06T10:56:01Z"}
> {"component":"entrypoint","error":"os: process already finished","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:255","func":"sigs.k8s.io/prow/pkg/entrypoint.gracefullyTerminate","level":"error","message":"Could not interrupt process after timeout","severity":"error","time":"2025-08-06T10:56:01Z"}
> {"component":"entrypoint","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:267","func":"sigs.k8s.io/prow/pkg/entrypoint.gracefullyTerminate","level":"error","message":"Process did not exit before 5m0s grace period","severity":"error","time":"2025-08-06T11:01:01Z"}
> {"component":"entrypoint","error":"os: process already finished","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:269","func":"sigs.k8s.io/prow/pkg/entrypoint.gracefullyTerminate","level":"error","message":"Could not kill process after grace period","severity":"error","time":"2025-08-06T11:01:01Z"}
> ```
> 
> Might the new loop have caused the tail process to restart, preventing the script from completing?

I addressed this in f4c0bb506162fd3023d088e432b9cdc3cec205c3 commit.

The while loop now runs in a separate process group created by `setsid` program. This allows us to send SIGTERM to the entire process group using `kill -TERM -- -$tail_pgid_leader` from the cleanup(). This makes sure that all child background processes (such as gcloud) are properly terminated when the main script exits. If `setsid` is not used, and the loop runs in a regular sub-shell, those child processes can continue running in the background. This prevents the container from shutting down properly, causing the Prow job to hang and eventually reach the 4-hour timeout.

PR #309, opened 2025-08-12
DBID Logger enhanced to handle extra spaces in the DBID output

sahsagar-google:
The updated playbook Sets Heading off and sql-trims the DBID. This new step effectively removes all leading and trailing whitespace from each line of the SQL query result before the DBID is extracted, preventing failures caused by inconsistent formatting.

### DIFF
--- a/roles/dbid-logger/tasks/main.yml
+++ b/roles/dbid-logger/tasks/main.yml
@@ -35,7 +35,7 @@
 
 - name: Extract DBID from query result
   ansible.builtin.set_fact:
-    oracle_dbid: "{{ dbid_query_result.stdout | trim if not dbid_query_result.failed else '' }}"
+    oracle_dbid: "{{ dbid_query_result.stdout | trim }}"
 
 - name: Output DBID
   ansible.builtin.debug:


roles/dbid-logger/tasks/main.yml
```diff
@@ -17,7 +17,9 @@
   ansible.builtin.shell: |
     set -o pipefail
     sqlplus -s / as sysdba <<EOF
-    SELECT dbid FROM v\$database;
+    SET HEADING OFF
+    SET FEEDBACK OFF
+    SELECT TRIM(dbid) FROM v\$database;
```
mfielding, line 22: a trim() of an implicit cast of a number has no effect.  The leading space is caused by sql*plus's formatting.

roles/dbid-logger/tasks/main.yml
```diff
@@ -17,7 +17,9 @@
   ansible.builtin.shell: |
     set -o pipefail
     sqlplus -s / as sysdba <<EOF
-    SELECT dbid FROM v\$database;
+    SET HEADING OFF
```
mfielding, line 20: Suggested wrapper text (internal link):  http://shortn/_ojm168852N

Notably, right now we ignore SQL errors.  We could add a "whenever sqlerror" clause, but do we actually want DB creation to fail if we couldn't get the DBID?  Perhaps combining whenever sqlerror with ignore_errors would propagate the failure in structured logging.

roles/dbid-logger/tasks/main.yml
```diff
@@ -30,10 +32,10 @@
 
 - name: Extract DBID from query result
   ansible.builtin.set_fact:
-    oracle_dbid: "{{ dbid_query_result.stdout_lines | select('match', '^\\d+$') | first | default('') | trim }}"
```
mfielding, line 33: It still doesn't hurt to trim here, especially in terms of trailing spaces.

mfielding: Also, all the sql*plus formatting still doesn't _guarantee_ that there won't be trailing whitespace.  Just to be conservative, consider a regex like `^\\w*\\d+\\w*$` that will match leading and trailing whitespace (that the trim command will then strip out).

roles/dbid-logger/tasks/main.yml
```diff
@@ -28,14 +31,13 @@
     PATH: "{{ oracle_home }}/bin:{{ ansible_env.PATH }}"
   register: dbid_query_result
   changed_when: false
-  failed_when: "'ORA-' in dbid_query_result.stderr"
 
-- name: Extract DBID from query result
+- name: Extract DBID and handle errors
   ansible.builtin.set_fact:
-    oracle_dbid: "{{ dbid_query_result.stdout_lines | select('match', '^\\d+$') | first | default('') }}"
+    oracle_dbid: "{{ dbid_query_result.stdout | trim if 'ORA-' not in dbid_query_result.stderr else '' }}"
```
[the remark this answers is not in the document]
sahsagar-google: Ack the first half.
For the second, won't it still be useful when a consumer goes looking into the logs for the DBID and could at least find an error instead?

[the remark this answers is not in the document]
sahsagar-google: After doing a test run, my reply stands corrected 
In this scenario, ignore_errors is a winner since besides just logging, it also lets the playbook continue but with awareness of the failure 
https://source.corp.google.com/piper///depot/google3/experimental/users/sagarsahai/oracle-toolkit/ansible-playbooks/failedwhenvsignoreerrors/README

roles/dbid-logger/tasks/main.yml
```diff
@@ -25,15 +31,13 @@
...
 
 - name: Extract DBID from query result
   ansible.builtin.set_fact:
-    oracle_dbid: "{{ dbid_query_result.stdout_lines | select('match', '^\\d+$') | first | default('') | trim }}"
+    oracle_dbid: "{{ dbid_query_result.stdout | trim if not dbid_query_result.failed else '' }}"
 
 - name: Output DBID
   ansible.builtin.debug:
-    msg: "ORACLE_DBID: {{ oracle_dbid }}"
+    msg: "{{ oracle_dbid if oracle_dbid else dbid_query_result.stderr | trim }}"
```
mfielding, line 42: Let's just output "{{ oracle_dbid | trim }}", so that a failure will produce and empty string.

sahsagar-google: You have a valid point, but the error message will be lost. 
Could we do the following?
    msg: "{{ oracle_dbid if oracle_dbid else dbid_query_result.stderr | trim }}"

mfielding: The error message will still be available in two places:
1. The structured logs for the previous step (as stderr)
2. The actual output from the previous step, as we use `ignore_errors` which outputs the full error dump.

mfielding (approved):

PR #321, opened 2025-08-26
Refactor the GCS download/transfer roles into a more logical and better flowing set of tasks

simonpane:
## Change Description:

Re-organize tasks and task files related to GCS procurement of media to provide a more logical, less redundant, and easier to follow set of organized steps (tasks).

## Solution Overview:

> **IMPORTANT NOTE TO REVIEWERS:** Might be easier to first review by looking at the re-factored task files as a whole vs looking at the differences/comparison view as the changes are significant

The existing tasks for downloading the software media from GCS storage buckets are split between two, substantially similar, task files: one (`gcstransfer.yml`) that transfers the files through the Ansible control node and another (`gcsdirect.yml`) that directly downloads from the swlib if the gcloud utility is available on the managed host.

Regardless of which task file is run, the steps are confusing and redundant. Specifically:

- There are many `shell` module tasks that effectively perform the same activities on the managed host.
- There are numerous interwoven `when` keys and jinja2 filters that are exactly replicated between the two files.
- If we were to add another download mechanism (i.e. from a private repository via URL), then we would replicate the above mentioned redundancy a third time.

Solution in this change is conceptually:

1. Build three lists of software to procure: one for the OPatch file (as it needs to be treated slightly differently as will be explained), one for base software (RDBMS and if required GI) which includes a primary file name and an alternate, and finally one for required patches. The content of each list will vary based on the chosen Oracle version and type of installation (i.e. full or `--no-patch` or `--ora-release base`).
2. Procure all of the files in one step per list: OPatch files are "always" re-downloaded (to ensure that the latest version is present/used), base software could be sourced from either the file name or alternate name, and patches are simply sourced by their name. The latter two, only re-sourced if they don't already exist.

Step 1 is performed in a new task file called `build_file_list.yml` that can be used under multiple conditions (such as direct download, or transfer through the control node). Then the `gcsdirect.yml` and `gcstransfer.yml` files are significantly simpler to follow and more organized with just three core tasks for the various procurement logic (with a fourth, simple, `set_fact` module in the latter) handling step 2.

> **NOTE:** Free edition software procurement is still treated separately via an additional and **unchanged** task in each file - this could possibly be simplified further in a separate and future PR.

## Test Results:

Test runs of various permutations with `verbosity: 1` for the `build_file_list | Display files lists` task:

- [Full typical install with Oracle 18c](https://gist.github.com/simonpane/19bed308fae316ccd881c9e6c989f1dc)
- [Full typical install with Oracle 21c](https://gist.github.com/simonpane/085f0c4dd5b49461e5eb046164be23a0)
- [Full typical install with Oracle 19c](https://gist.github.com/simonpane/521c3a273b779777fb81d9a9e04597be)
- [Full day-2 patched install with Oracle 19c](https://gist.github.com/simonpane/85823a105f5600753fb20db1a03219ad)
- [Full typical install with Oracle 19c using the "base" option](https://gist.github.com/simonpane/2935d7ef95751aa319a7df5d5024fa1d)
- [Free Edition install](https://gist.github.com/simonpane/aff9ea2ae10703dce18d4797a63ba921)
- [Full typical install with Oracle 19c using Ansible 2.9](https://gist.github.com/simonpane/3361b7393d0818b8ff8223bd6dff1297)

### DIFF
--- a/roles/swlib/tasks/gcsdirect.yml
+++ b/roles/swlib/tasks/gcsdirect.yml
@@ -13,10 +13,10 @@
 # limitations under the License.
 
 ---
-- name: gcsdirect | Copy OPatch (always)
+- name: gcsdirect | Copy OPatch
   shell: |
     set -o pipefail
-    gcloud storage cp gs://{{ swlib_mount_src }}/{{ item }} "{{ swlib_path }}/{{ item }}"
+    gcloud storage cp "gs://{{ swlib_mount_src }}/{{ item }}" "{{ swlib_path }}/{{ item }}"
   register: shell_result
   args:
     executable: /bin/bash
@@ -26,13 +26,13 @@
 - name: gcsdirect | Copy base software files using file_name or alt_name
   shell: |
     set -o pipefail
-    if gcloud storage ls gs://{{ swlib_mount_src }}/{{ item.file_name }} >/dev/null 2>&1; then
+    if gcloud storage ls "gs://{{ swlib_mount_src }}/{{ item.file_name }}" >/dev/null 2>&1; then
       f_name="{{ item.file_name }}"
     else
       f_name="{{ item.alt_name | default('x') }}"
     fi
     if ! unzip -l "{{ swlib_path }}/${f_name}"; then
-      gcloud storage cp gs://{{ swlib_mount_src }}/${f_name} "{{ swlib_path }}/{{ item.file_name }}"
+      gcloud storage cp "gs://{{ swlib_mount_src }}/${f_name}" "{{ swlib_path }}/{{ item.file_name }}"
     fi
   register: shell_result
   args:
@@ -45,15 +45,13 @@
     set -o pipefail
     if ! unzip -l "{{ swlib_path }}/{{ item }}"
       then
-        gcloud storage cp gs://{{ swlib_mount_src }}/{{ item }} "{{ swlib_path }}/{{ item }}"
+        gcloud storage cp "gs://{{ swlib_mount_src }}/{{ item }}" "{{ swlib_path }}/{{ item }}"
     fi
   register: shell_result
   args:
     executable: /bin/bash
   with_items: "{{ patch_file_list | unique }}"
   changed_when: "'Date    Time    Name' not in shell_result.stdout"
-
-################################################################################
 
 - name: gcsdirect | Download or copy RPM software from GCS to target instance
   shell: |
@@ -63,7 +61,7 @@
       if [[ "{{ f.name }}" == https://*oracle.com/* || "{{ f.name }}" == http://*oracle.com/* ]]; then
         curl --location --silent --connect-timeout 20 --fail -o "{{ swlib_path }}/{{ f.name | basename }}" "{{ f.name }}" && echo "File downloaded"
       else
-        gcloud storage cp gs://{{ swlib_mount_src }}/{{ f.name }} "{{ swlib_path }}/{{ f.name }}" && echo "File copied"
+        gcloud storage cp "gs://{{ swlib_mount_src }}/{{ f.name }}" "{{ swlib_path }}/{{ f.name }}" && echo "File copied"
       fi
     fi
     {% endfor %}
--- a/roles/swlib/tasks/gcstransfer.yml
+++ b/roles/swlib/tasks/gcstransfer.yml
@@ -17,10 +17,10 @@
   set_fact:
     ssh_cmd: "ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }}"
 
-- name: gcstransfer | Transfer OPatch (always)
+- name: gcstransfer | Transfer OPatch
   shell: |
     set -o pipefail
-    gcloud storage cat gs://{{ swlib_mount_src }}/{{ item }} | {{ ssh_cmd }} "cat > {{ swlib_path }}/{{ item }}"
+    gcloud storage cat "gs://{{ swlib_mount_src }}/{{ item }}" | {{ ssh_cmd }} "cat > \"{{ swlib_path }}/{{ item }}\""
   register: shell_result
   args:
     executable: /bin/bash
@@ -31,13 +31,13 @@
 - name: gcstransfer | Transfer base software files using file_name or alt_name
   shell: |
     set -o pipefail
-    if gcloud storage ls gs://{{ swlib_mount_src }}/{{ item.file_name }} >/dev/null 2>&1; then
+    if gcloud storage ls "gs://{{ swlib_mount_src }}/{{ item.file_name }}" >/dev/null 2>&1; then
       f_name="{{ item.file_name }}"
     else
       f_name="{{ item.alt_name | default('x') }}"
     fi
-    if ! {{ ssh_cmd }} unzip -l "{{ swlib_path }}/${f_name}"; then
-      gcloud storage cat gs://"{{ swlib_mount_src }}"/"${f_name}" | {{ ssh_cmd }} "cat > {{ swlib_path }}/{{ item.file_name }}"
+    if ! {{ ssh_cmd }} "unzip -l \"{{ swlib_path }}/${f_name}\""; then
+      gcloud storage cat "gs://{{ swlib_mount_src }}/${f_name}" | {{ ssh_cmd }} "cat > \"{{ swlib_path }}/{{ item.file_name }}\""
     fi
   register: shell_result
   args:
@@ -49,8 +49,8 @@
 - name: gcstransfer | Transfer patch files
   shell: |
     set -o pipefail
-    if ! {{ ssh_cmd }} unzip -l "{{ swlib_path }}/{{ item }}"; then
-      gcloud storage cat gs://{{ swlib_mount_src }}/{{ item }} | {{ ssh_cmd }} "cat > {{ swlib_path }}/{{ item }}"
+    if ! {{ ssh_cmd }} "unzip -l \"{{ swlib_path }}/{{ item }}\""; then
+      gcloud storage cat "gs://{{ swlib_mount_src }}/{{ item }}" | {{ ssh_cmd }} "cat > \"{{ swlib_path }}/{{ item }}\""
     fi
   register: shell_result
   args:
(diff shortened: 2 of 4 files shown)


simonpane:
@mfielding - this PR includes a new suggested task file.  Please feel free to comment on the name of the new file and provide an alternative file name if you would like?  Thanks.

roles/swlib/tasks/gcsdirect.yml
```diff
@@ -13,111 +13,34 @@
 # limitations under the License.
 
 ---
-# files with alt_name are stored as name for simplicity
-- name: gcsdirect | Copy GI software from GCS to target instance
+- name: gcsdirect | Copy OPatch (always)
```
mfielding, line 16: Can we remove the (always) in the name?  If using a base release or free edition, `opatch_file_list` will be empty and the step intentionally skipped.

roles/swlib/tasks/gcsdirect.yml
```diff
@@ -13,111 +13,34 @@
...
+      f_name="{{ item.file_name }}"
     else
-      f_name="{{ f.alt_name | default('x') }}"
+      f_name="{{ item.alt_name | default('x') }}"
     fi
-    if ! unzip -l "{{ swlib_path }}/{{ f.name }}"
-      then
-        gcloud storage cp gs://{{ swlib_mount_src }}/${f_name} "{{ swlib_path }}/{{ f.name }}"
+    if ! unzip -l "{{ swlib_path }}/${f_name}"; then
+      gcloud storage cp gs://{{ swlib_mount_src }}/${f_name} "{{ swlib_path }}/{{ item.file_name }}"
```
mfielding, line 35: If using the alt_name, we're actually now saving it using file_name;  is this an intentional change?

mfielding: Can we please quote the first file name too?

roles/swlib/tasks/gcsdirect.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2025 Google LLC
```
mfielding, line 1: nit: no need to update this for existing files: https://opensource.google/documentation/reference/copyright

roles/swlib/tasks/gcsdirect.yml
```diff
@@ -127,10 +50,11 @@
   register: shell_result
   args:
     executable: /bin/bash
-  with_items: "{{ rdbms_patch_files }}"
-  when: rdbms_patch_files is defined and rdbms_patch_files | length > 0
+  with_items: "{{ patch_file_list | unique }}"
   changed_when: "'Date    Time    Name' not in shell_result.stdout"
 
+################################################################################
```
mfielding, line 56: Looks like a debug comment here?

roles/swlib/tasks/gcstransfer.yml
```diff
@@ -13,137 +13,53 @@
...
-    executable: /bin/bash
-  with_items: "{{ gi_patches }}"
-  when:
-    - gi_install
-    - item.release == oracle_rel
-  changed_when: "'Date    Time    Name' not in shell_result.stdout"
-  delegate_to: 127.0.0.1
+- name: gcstransfer | Set the Ansible ssh command for use in subsequent file tasks
+  set_fact:
+    ssh_cmd: "ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }}"
```
mfielding, line 18: Would we get a benefit by setting this globally?  It seems like the kind of thing we might use in other playbooks too.

roles/swlib/tasks/gcstransfer.yml
```diff
@@ -13,137 +13,53 @@
...
-# overwrite OPatch zipfile on database host from gcs every time that may be cached currently in {{ swlib_path }}
-# reason to overwrite is that the OPatch zipfile in MOS is named the same, example: p6880880_190000_Linux-x86-64.zip
-# this change downloads OPatch from GCS bucket every time with negligible overhead as OPatch zip file is typically ~120M
-- name: gcstransfer | Transfer OPatch update files from GCS to target instance
+- name: gcstransfer | Transfer OPatch (always)
   shell: |
     set -o pipefail
-    gcloud storage cat gs://{{ swlib_mount_src }}/{{ item.patchfile }} | ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} \
-      {{ ansible_ssh_user }}@{{ ansible_ssh_host }} "cat > {{ swlib_path }}/{{ item.patchfile }}"
+    gcloud storage cat gs://{{ swlib_mount_src }}/{{ item }} | {{ ssh_cmd }} "cat > {{ swlib_path }}/{{ item }}"
```
mfielding, line 23: (Again, pre-existing issue) Can you quote the actual file name too?  might require some escaping of quotes.

roles/swlib/tasks/gcstransfer.yml
```diff
@@ -13,137 +13,53 @@
...
+- name: gcstransfer | Transfer patch files
   shell: |
     set -o pipefail
-    if ! ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} {{ ansible_ssh_user }}@{{ ansible_ssh_host }} \
-      "unzip -l {{ swlib_path }}/{{ item }}"
-      then
-        gcloud storage cat gs://{{ swlib_mount_src }}/{{ item }} | ssh -i {{ ansible_ssh_private_key_file }} {{ ansible_ssh_extra_args }} \
-          {{ ansible_ssh_user }}@{{ ansible_ssh_host }} "cat > {{ swlib_path }}/{{ item }}"
+    if ! {{ ssh_cmd }} unzip -l "{{ swlib_path }}/{{ item }}"; then
+      gcloud storage cat gs://{{ swlib_mount_src }}/{{ item }} | {{ ssh_cmd }} "cat > {{ swlib_path }}/{{ item }}"
```
mfielding, line 53: Similarly here with the quoting.

roles/swlib/tasks/main.yml
```diff
@@ -40,10 +40,14 @@
   register: gcloud_found
   changed_when: false
 
+- name: swlib | Determine required installation files
+  include_tasks: build_file_list.yml
+  when: swlib_mount_type not in ["gcsfuse","nfs"]
```
mfielding, line 45: For consistency with the subsequent steps, should we do something like `swlib_mount_type in ["gcsdirect", "gcs"]` ?

roles/swlib/tasks/main.yml
```diff
@@ -40,10 +40,14 @@
...
+
 - name: swlib | Copy files from GCS directly on the Managed Host instance
   include_tasks: gcsdirect.yml
-  when: (swlib_mount_type == "gcsdirect" or swlib_mount_type == "gcs") and (gcloud_found is defined and gcloud_found.stdout == "0")
+  when: swlib_mount_type in ["gcsdirect", "gcs"] and gcloud_found is defined and gcloud_found.stdout == "0"
 
 - name: swlib | Use the Ansible Control Node to transfer files from GCS to the Managed Host instance
   include_tasks: gcstransfer.yml
-  when: swlib_mount_type == "gcstransfer" or gcloud_found is not defined or (gcloud_found is defined and gcloud_found.stdout != "0")
+  when: swlib_mount_type == "gcstransfer" or (swlib_mount_type in ["gcsdirect", "gcs"] and gcloud_found.stdout | default("1") != "0")
```
mfielding, line 53: Can we stick with the `gcloud_found is defined and gcloud_found.stdout == "0"` type syntax of the previous check?

mfielding (commented):
Yeah, with the recent changes, this type of refactoring definitely makes sense.  I've added a few minor comments, but relying on your tests to make sure that it actually works end-to-end.

roles/swlib/tasks/main.yml
simonpane: @mfielding , I did this intentionally as I think that we might provide other options in the future so the inclusion list will be longer than the exclusion list.  For example `in ["gcs","gcsdirect","gcstransfer","url"]` .  That last option doesn't exist currently but conceptually could in the future.  And possibly could expand into other permutations of URL (i.e. based on authentication type).

Conversely, the exclusion list I think will be rather small and somewhat static.  Even if we ran this task-file by mistake, it's very fast (as it's just a series of `set_fact` modules and jinja2 filters) and benign. 

Based on that explanation, please let me know if you'd still like me to reverse the logic.  Thanks.

roles/swlib/tasks/gcsdirect.yml
mfielding: No worries, this is a small nit anyway.

roles/swlib/tasks/main.yml
mfielding: The only playbooks that can consume the build_file_list output today are gcs, gcsdirect, and gcstransfer.  I still think it reads easier to list them in the same form here as for the two following steps.  If we add URL in the future, we'd need to add it here too, yes.

roles/swlib/tasks/gcsdirect.yml
simonpane: Removed in commit https://github.com/google/oracle-toolkit/pull/321/commits/5e67b006f3e1a22901f693ebc4b48a4ed9877674.

roles/swlib/tasks/gcstransfer.yml
simonpane: Quoting adjusted in commit https://github.com/google/oracle-toolkit/pull/321/commits/5e67b006f3e1a22901f693ebc4b48a4ed9877674.  If you'd like me to change it to use single quotes on the remote command so that we don't have to escape the inner quotes, then just let me know.

roles/swlib/tasks/main.yml
simonpane: Reverted back to the old boolean logic.

roles/swlib/tasks/main.yml
simonpane: Changed to an inclusion list accordingly in commit https://github.com/google/oracle-toolkit/commit/5e67b006f3e1a22901f693ebc4b48a4ed9877674.

roles/swlib/tasks/gcstransfer.yml
mfielding: In that case let's leave as-is;  we can refactor later.

mfielding (approved):

PR #324, opened 2025-08-28
Add `whenever sqlerror exit failure` to all sqlplus blocks

simonpane:
## Change Description:

Add `whenever sqlerror exit sql.sqlcode` to all sqlplus blocks.

## Solution Overview:

SQL*Plus returns a `0` return code in most situations when SQL or PL/SQL blocks fail.  This results in Ansible usually reporting the associated task as a "**success**" when in fact it may have "**failed**" at the SQL level.  Consequently, SQL or PL/SQL failures are not caught or realized.

### SQL*Plus Return Code Examples:

HERESTRING example:

```plaintext
$ sqlplus -s / as sysdba <<< "select * from bogus;"
select * from bogus
              *
ERROR at line 1:
ORA-00942: table or view "SYS"."BOGUS" does not exist
Help: https://docs.oracle.com/error-help/db/ora-00942/


$ echo $?
0
$
```

HEREDOC example:

```plaintext
$ sqlplus -s / as sysdba <<EOF
select * from bogus;
EOF

select * from bogus
              *
ERROR at line 1:
ORA-00942: table or view "SYS"."BOGUS" does not exist
Help: https://docs.oracle.com/error-help/db/ora-00942/


$ echo $?
0
$
```

PL/SQL block in HEREDOC example:

```plaintext
$ sqlplus -s / as sysdba <<EOF
BEGIN
   bogus_function;
END;
/
EOF
   bogus_function;
   *
ERROR at line 2:
ORA-06550: line 2, column 4:
PLS-00201: identifier 'BOGUS_FUNCTION' must be declared
ORA-06550: line 2, column 4:
PL/SQL: Statement ignored
Help: https://docs.oracle.com/error-help/db/ora-06550/


$ echo $?
0
$
```

**Conclusion is that SQL and/or PL/SQL in HEREDOCs and HERESTRINGs are not returning non-zero return codes and consequently Ansible tasks may falsely show a "success"**

## Test Results:

- [Full test of 19c EE installation with ASM storage](https://gist.github.com/simonpane/b714151bcf569199baf98af88ee08d7f)
- [Full test of 19c EE installation with FS storage](https://gist.github.com/simonpane/4ce1a54a6c2e9e44da5db6f08ff3771e)
- [Full test of 23ai Free Edition installation](https://gist.github.com/simonpane/d5be029742ca61bea7f1c4cab90bd229)

PR #326, opened 2025-08-28
Fix multitenant check when saving PDB state

mfielding:
The `pdbs-state` task only runs when `pdb_prefix` is defined.  But `pdb_prefix` is _always_ defined, with a default value of PDB.  This change uses `container_db` instead.  This is a string (not boolean) value set to "true" or "false".

This issue was discovered in #324 tests, when we turned on `whenever sqlerror` and discovered that this has been failing: https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/324/oracle-toolkit-install-data-guard-on-gcp/1961154540484956160


mfielding:
> Have you considered converting "container_db" to a boolean in the group_vars/all.yml file?

Done, along with stringifying the representations.

AlexBasinov (approved):

PR #327, opened 2025-08-29
Set pwd_gen_cmd conditionally based on ora-db-container flag

AlexBasinov:
Set 'pwd_gen_cmd' conditionally based on the value of 'ora-db-container' flag.

This PR fixes a bug where multitenant deployments fail when retrieving a password from Secret Manager. The pwd_gen_cmd variable was being set with only two password lines, but the dbca tool expects three (sys, system, and pdbadmin). 

Internal bug: b/441736294

Test results:

Random password+ CDB → [Logs](https://gist.github.com/AlexBasinov/7789235c73352a4bd751f952fee3dc8d)
Random password + non-CDB → [Logs](https://gist.github.com/AlexBasinov/2a1f239615a8fb64cb989eb8a61bf52e)
Password from Secret + CDB → [Logs](https://gist.github.com/AlexBasinov/211ae9e1ac06fcc5a68aa618d1566c85)
Password from Secret + non-CDB → [Logs](https://gist.github.com/AlexBasinov/38b2762347b3800cbba44a76605af9ae)


group_vars/all.yml
```diff
@@ -70,7 +70,7 @@ oracle_rel: "{{ lookup('env','ORA_RELEASE') | default('latest',true) }}"
 oracle_edition: "{{ lookup('env','ORA_EDITION') | default('EE',true) }}"
 cluster_type: "{{ lookup('env','CLUSTER_TYPE') | default('NONE',true) }}"
 
-container_db: "{{ lookup('env','ORA_DB_CONTAINER') | lower | default('false',true) }}"
+container_db: "{{ lookup('env','ORA_DB_CONTAINER') | lower | default('false',true) | bool }}"
```
mfielding, line 73: This change appears to negate the default:  `'false' | bool` in jinja evaluates to true.  

Consider the logic in https://github.com/google/oracle-toolkit/pull/326/commits/2b146c9788d694ab8e6fbb050efbefe0a1a3cb25#diff-7be6987d456ad921f461608b488a7593f037ba8fc6e43ed862c903e91109c00aL74 instead

```jinja
container_db: "{{ lookup('env','ORA_DB_CONTAINER') | default(false,true) | bool }}"
```

roles/db-create/tasks/main.yml
```diff
@@ -97,26 +97,36 @@
...
   tags: db-create
 
-- name: Generate command for password randomization (not 11.2)
+- name: Generate command for password randomization for CDB
   set_fact:
     pwd_gen_cmd: echo -e "$({{ pass_param }})\n$({{ pass_param }})\n$({{ pass_param }})"
   when:
-    - oracle_ver != "11.2.0.4.0"
     - db_password_secret | length == 0
+    - container_db | bool
```
mfielding, line 105: nit: if converting to a bool in `group_vars/all.yml`, it's probably safer not to explicitly cast to a bool here, to cause the evaluation to error out if a string somehow sneaks in.

mfielding (commented):
Oh, and there's another container_db reference in the code that turns into an implicit cast if the data type is changed.  #326 includes a fix.

PR #328, opened 2025-09-01
Take `oracle_edition` into consideration when building the RDBMS software file list

simonpane:
## Change Description:

Take `oracle_edition` into consideration when building the RDBMS software file list.

## Solution Overview:

Currently the RDBMS software list is not taking into consideration the Oracle edition (the `oracle_edition` Ansible variable).  And the only reason that it is currently working is based on:

- No overlap between the currently available Oracle versions/release Free edition and EE/SE2 editions
- Same software required for both EE and SE2

This change rectifies this by taking the `oracle_edition` into consideration when determining the software list.  Including consideration for when the edition information is provided, in the `rdbms_software` dict, as a key-value pair (as is the case for `FREE` edition) and a key-list (as is the case for the commercial editions).

## Test Results:

- [Full test run using Free edition](https://gist.github.com/simonpane/fb8652646e1627e6a4428ecea61bc4ac)
- [Full test run using 19c EE with Day 1 patching](https://gist.github.com/simonpane/6977ab26b85071e9174ae4f6964a21ae)
- [Full test run using 19c EE with Day 2 patching](https://gist.github.com/simonpane/9eecc09ebef86af894a31f02e81e8af9)

### DIFF
--- a/roles/swlib/tasks/build_file_list.yml
+++ b/roles/swlib/tasks/build_file_list.yml
@@ -42,8 +42,8 @@
 
 - name: build_file_list | Build list of RDBMS base software by edition
   set_fact:
-    rdbms_edition_sw: "{{ (rdbms_software | selectattr('edition', 'equalto', oracle_edition))
-                        + (rdbms_software | selectattr('edition', 'search', '\\b' + oracle_edition + '\\b'))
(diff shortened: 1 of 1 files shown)


mfielding (approved):
Nice catch. Good to fix this before it becomes a problem.

PR #330, opened 2025-09-02
Changes to support optionally procuring software from a URL

simonpane:
## Change Description:

Allow (undocumented) option to download software from a URL based location.

## Solution Overview:

Instead of using a GCS bucket, some customers may potentially want the ability to download the software from some internally staged, and URL accessed, private location or artifact repository. Such as `https://media.company.com`.

This change will detect if the `--ora-swlib-bucket` value provided begins with `http://` or `https://` and if so, will procure the software using the Ansible `uri` module.

Optional parameters to the `uri` module can be provided as as **OS environment variables** as required. Including: `URL_USERNAME`, `URL_PASSWORD` and `URL_HEADERS`. The latter must be in the form of properly formed JSON.

Example of setting the `URL_HEADERS` OS environment variable (in this example accessing a GCS bucket via URL simply for the purpose of demonstrating this functionality):

```bash
export TOKEN=$(curl -s "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google" | jq -r .access_token)
export URL_HEADERS="{\"Authorization\": \"Bearer ${TOKEN}\"}"
echo ${URL_HEADERS} | jq -r '.'
```

Then run the rest of the toolkit installation as normal, the only exception being the URL-based source location. For example:

```bash
./install-oracle.sh \
  --ora-version 19 \
  --ora-swlib-bucket https://storage.googleapis.com/pythian-gto-oracle-software/19c/ \
  --ora-asm-disks-json '[{"diskgroup":"DATA","disks":[{"blk_device":"/dev/disk/by-id/google-oracle-data-1","name":"DATA1"}]},{"diskgroup":"RECO","disks":[{"blk_device":"/dev/disk/by-id/google-oracle-reco-1" ,"name":"RECO1"}]}]' \
  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"}]' \
  --backup-dest "+RECO"
```

> **NOTE:** The other OS environment variables `URL_USERNAME` and `URL_PASSWORD` can be left undefined but are included in the Ansible tasks for their optional usage and advance-user reference.

## Test Results:

- [Free Edition install using Ansible 2.9](https://gist.github.com/simonpane/78a8bef6a07e2275b0cb64fd44ac0497)
- [Installation of 19c EE with Day 2 patching](https://gist.github.com/simonpane/ac1ac5bcba5297235142c5ee85c373bd)

### DIFF
--- a/apply-patch.sh
+++ b/apply-patch.sh
@@ -32,7 +32,10 @@ ORA_RELEASE="${ORA_RELEASE}"
 ORA_RELEASE_PARAM=""
 
 ORA_SWLIB_BUCKET="${ORA_SWLIB_BUCKET}"
-ORA_SWLIB_BUCKET_PARAM="^gs://.+"
+ORA_SWLIB_BUCKET_PARAM="^(gs:\/\/|https?:\/\/)"
+
+ORA_SWLIB_TYPE="${ORA_SWLIB_TYPE:-GCS}"
+ORA_SWLIB_TYPE_PARAM="^(\"\"|GCS|GCSFUSE|NFS|GCSDIRECT|GCSTRANSFER)$"
 
 ORA_SWLIB_PATH="${ORA_SWLIB_PATH:-/u01/swlib}"
 ORA_SWLIB_PATH_PARAM="^/.*"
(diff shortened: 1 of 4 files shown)


mfielding:
Giving this some thought.  I like the simple and straightforward approach, but am just wondering what can be reasonably funcitonal but consistent with a secure-by-default configuration.

PR #331, opened 2025-09-02
Changes to support installing Free edition on EL9 with new `os_version` key-value in `rdbms_software`

simonpane:
## Change Description:

Add support for installing Free edition on Enterprise Linux 9. And the dependent change of including an `os_version` key-value in the `rdbms_software` list of dicts to allow OS major version specific installations.

This is an **initial change** for allowing OS major version specific installations, with a focus on installing Free Edition on Enterprise Linux 9.

## Solution Overview:

Some installation media will be OS major version specific and other media will work with multiple OS versions. Consequently, a new `os_version` key-value that can either be a string value or a list is added to the `rdbms_software` list of dictionaries.

The `build_file_list.yml` task file now takes this into consideration when determining the required software to procure and install. (For Ansible 2.9 support, this must be done through a separate `set_fact` module task.)

Other aspects of OS major version an EL9 support **to come in complementary but separate and forthcoming PR** that will include:

- Changes required to **install 19c EE/SE2 on Enterprise Linux 9** (hence `rdbms_software` updates for non-Free edition currently do not list EL 9)
- Enhanced OS major version and other parameters compatibility checks

## Test Results:

Since the `rdbms_software` list is adjusted, tests to install Free edition on OL9 as well as to install EE against OL8 (a regression test essentially) were performed:

- [Free Edition installation using Oracle Linux 9](https://gist.github.com/simonpane/d562b157d94bd30e6278c710e65bd2c5)
- [19c EE installation using Oracle Linux 8](https://gist.github.com/simonpane/be09b6bebf3a30622635891846199a15)

Further, to ensure that the OS major version matched files were properly selected when installing 19c EE on OL8, added a dummy OL9 dict to `rdbms_software` to ensure that it was not used by the toolkit (this dummy entry was used during testing only and is not included in this PR):

```yaml
  - name: 19c_base
    version: 19.3.0.0.0
    edition:
      - EE
      - SE2
    os_version: 9
    files:
      - { name: "foobar.zip", sha256sum: "foobar", md5sum: "foobar==",
          alt_name: "foobar_db_home.zip", alt_sha256sum: "foobar", alt_md5sum: "foobar==" }
```

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -155,48 +155,63 @@ rdbms_software:
   - name: 23ai_free_23_9
     version: 23.9.0.25.07
     edition: FREE
+    os_version: 9
+    files:
+      - { name: "https://yum.oracle.com/repo/OracleLinux/OL9/appstream/x86_64/getPackage/oracle-database-preinstall-23ai-1.0-2.el9.x86_64.rpm", sha256sum: "aa7bc3a62f4118cc8e02ece2f67ddd276b2256833e4d66f939725b2ef22bebf9", md5sum: "6opRa2Ug+YMZaITFabPt0g=="}
+      - { name: "https://download.oracle.com/otn-pub/otn_software/db-free/oracle-database-free-23ai-23.9-1.el9.x86_64.rpm", sha256sum: "35a5b2e4065747eea3258d4f0c8d9a6e5440a818945da183fc631750cce4d999", md5sum: "HiohdiKUMAHK/ejTxJR72w=="}
+  - name: 23ai_free_23_9
+    version: 23.9.0.25.07
+    edition: FREE
+    os_version: 8
     files:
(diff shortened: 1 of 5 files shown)


simonpane:
> A quick thought: the OS version requirements seem to be specific to the RPM files we use for free edition.
> 
> Is it possible to restrict os_version to just RPMs and not base ZIP files?
> 
> For example, even if Oracle 12c is only supported on EL7, I don't think this is the right place to enforce that requirement.

Understood and I've adjusted accordingly in commit https://github.com/google/oracle-toolkit/pull/331/commits/0bd19af67c026e3eb7eb26e01e25aaadb109d9e4.  Removed the `os_version` variables from the non-Free Edition installs and refactored the jinja2 filters to make it optional.

It's actually a pretty flexible implementation - the `os_version` variable in the dict is optional and it can be a single value or a list.  As it's possible in the future that Oracle could release RPMs that apply to multiple Linux versions.  And similarly, while unlikely, they could theoretically release OS specific zip files.  This change supports all of those possibilities.

Re-tested with Free Edition on OL8, Free Edition on OL9, 19c EE without patching and 19c EE with day-1 patching.  FYI @mfielding

PR #336, opened 2025-09-05
Refactor presubmit tests to support free edition

mfielding:
We're currently missing presubmit test coverage for all three of 23c, free edition, and filesystem installs.  This change adds a new Free Edition presubmit test.

To avoid too much code duplication with the existing data guard tests, we move the shared functionality into a new `infra-manager-lib.sh` that is sourced from each. And yes, software engineers will roll their eyes at the use of global variables here.

Some smaller updates while refactoring the shell:
* More explicit error checking and exiting
* In the case of a Terraform error, run `gcloud infra-manager deployments describe` to fetch the actual error message before tearing down the whole deployment (and losing the error)
* Do template replacement out-of-place into a temporary file, so that the presubmit checks can be safely re-run with different parameters

Sample output:  https://gist.github.com/mfielding/b8ee3e3699d2be67dcb59fa3d3afed02

Shellcheck runs cleanly:

```
mfielding@nari:~/git/oracle-toolkit$ shellcheck -x -a presubmit_tests/free-edition-on-gcp.sh
mfielding@nari:~/git/oracle-toolkit$ shellcheck -x -a presubmit_tests/data-guard-on-gcp.sh
```


presubmit_tests/infra-manager-lib.sh
```diff
@@ -0,0 +1,178 @@
...
+  # This prevents one test from deleting the file while it's still in use by another concurrently running test.
+  # For available Prow-injected environment variables, see:
+  # https://docs.prow.k8s.io/docs/jobs/#job-environment-variables
+  toolkit_zip_file_name="oracle-toolkit-${BUILD_ID}.zip"
+  project_id="$(curl -fsS -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/project/project-id")" || {
+    echo "Error: Failed to retrieve project ID" >&2
+    exit 1
+  }
+  gcs_source="${gcs_bucket}/${toolkit_zip_file_name}"
+  deployment_id="projects/${project_id}/locations/${location}/deployments/${deployment_name}"
```
AlexBasinov, line 36: Should the setup_vars() verify that the caller set the ${deployment_name}?

AlexBasinov (approved):

PR #352, opened 2025-09-15
Optimize tasks related to managing the firewall (`firewalld`) service

simonpane:
## Change Description:

Optimize tasks related to managing the firewall (`firewalld`) service

## Solution Overview:

Specific optimizations include:

1. Removing the redundant and un-used `disable_firewall` variable declaration in the `defaults/main.yml` file. The `group_vars/all.yml` file has precedence over any role's `defaults/main.yml` file meaning the role specific setting was never being used. Which can be confusing for toolkit maintainers. The declaration was also in a marked as "non-customizable variables" section of the `group_vars/all.yml` file - moved to a different section of that same file for clarity. (NOTE: defining in `group_vars/all.yml` is probably the most appropriate location since this variable is used in multiple, independent tasks, from different playbooks.)

2. Removing legacy references to Enterprise Linux 6 and the old `iptables` command. Especially relevant since the toolkit verifies that the minimum OS version is at least EL 7.3 via the `os_min_supported_version` variable. The firewall service name is still left as a variable (rather than using a hard-coded `firewalld` value in the actual tasks) for future flexibility, for example, should the service name change or an alternate firewall program be used.

3. Updated logic for checking whether the firewall (`firewalld`) service is present on the system. Using the `package` module, especially with the `disablerepo: "*"` option as the playbook currently does, is not optimal, is inefficient, and could possibly have undesirable side-effects. The alternative of checking on a file such as `/etc/systemd/system/firewalld.service` via the `stat` module to check for a service's existence is also not reliable as the location/file name could change. Therefore using the `service_facts` Ansible module is probably optimal and allows a subsequent task to only stop/disable the firewall service if found.

## Test Results:

When service is present and is running:

```plaintext
...
TASK [ora-host : Gather service facts] ******************************************************************************************************************************************************************
ok: [10.2.80.121]

TASK [ora-host : Stop and disable firewall if present] **************************************************************************************************************************************************
changed: [10.2.80.121]

...
```

When service is present and not running:

```plaintext
...
TASK [ora-host : Gather service facts] ******************************************************************************************************************************************************************
ok: [10.2.80.121]

TASK [ora-host : Stop and disable firewall if present] **************************************************************************************************************************************************
ok: [10.2.80.121]

...
```

When service is not present (uninstalled `firewalld` on managed host):

```plaintext
...
TASK [ora-host : Gather service facts] ******************************************************************************************************************************************************************
ok: [10.2.80.121]

TASK [ora-host : Disable selinux (permanently)] *********************************************************************************************************************************************************
ok: [10.2.80.121]

...
```

### DIFF
--- a/roles/ora-host/tasks/main.yml
+++ b/roles/ora-host/tasks/main.yml
@@ -26,6 +26,8 @@
     state: stopped
     enabled: false
   when:
+    - disable_firewall | bool
+    - ansible_os_family == 'RedHat'
     - ansible_facts.services is defined
     - firewall_service + ".service" in ansible_facts.services
   tags: firewall


roles/ora-host/tasks/main.yml
```diff
@@ -13,26 +13,21 @@
...
+- name: Stop and disable firewall if present
   service:
     name: "{{ firewall_service }}"
     state: stopped
     enabled: false
   when:
-    - disable_firewall | bool
-    - ansible_os_family == 'RedHat'
-    - firewall.results
+    - ansible_facts.services is defined
```
mfielding, line 29: nit: reading this, it isn't obvious that this command only runs if disable_firewall is set.  Would you mind making this dependent on disable_firewall and ansible_os_family as before for readability, even if it's an extra line here?

mfielding (approved):

PR #357, opened 2025-09-17
Collect usage metrics for Terraform deployments via workload agent

sahsagar-google:
Adds new Ansible tasks to find the google-cloud-workload-agent executable and run the logusage command.

These logging tasks are made conditional and will only run if the agent installation is requested. This is done by checking the ansible_env.INSTALL_WORKLOAD_AGENT variable, which is already exported by the install-oracle.sh script.

Buganizer: b/427723342

### DIFF
--- a/roles/workload-agent/tasks/main.yml
+++ b/roles/workload-agent/tasks/main.yml
@@ -17,6 +17,32 @@
   include_tasks:
     file: install.yml
   tags: workload-agent
+  
+- name: Find the exact path of the installed agent executable
+  ansible.builtin.shell:
+    # Query the RPM database for the package and grep for the main executable
+    # The '$' at the end ensures we get the main binary, not a helper script
+    cmd: "rpm -ql google-cloud-workload-agent | grep '/bin/google-cloud-workload-agent$'"
+  register: agent_exe_path
+  changed_when: false
+  ignore_errors: true
+  tags: workload-agent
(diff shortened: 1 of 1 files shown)


roles/workload-agent/tasks/main.yml
```diff
@@ -17,6 +17,32 @@
   include_tasks:
     file: install.yml
   tags: workload-agent
+  
+- name: Find the exact path of the installed agent executable
```
mfielding, line 21: If coming from the rpm, the workload agent executable always gets installed into `/usr/bin/google_cloud_workload_agent`; I think we'd be fine using the default path, which will include `/usr/bin`.

roles/workload-agent/tasks/main.yml
```diff
@@ -17,6 +17,32 @@
...
+  when: ansible_env.INSTALL_WORKLOAD_AGENT is defined and ansible_env.INSTALL_WORKLOAD_AGENT == "true"
+
+- name: Log workload agent installation
+  ansible.builtin.shell:
+    cmd: "{{ agent_exe_path.stdout }} logusage -s ACTION -a 5"
+  register: log_result
+  changed_when: false
+  ignore_errors: true
+  tags: workload-agent
+  when:
```
mfielding, line 39: I don't think we need any of this:  this particular task follows an agent install.

mfielding (approved):

PR #360, opened 2025-09-25
Improve password checks

mfielding:
Update password checks to be more in line with the actual requirments for SYS passwords as enforced by DBCA.  This is a higher standard than the DEFAULT password policy.  This change will increase the complexity standard for the Oracle metrics user, which formerly weren't validated for complexity, only invalid characters.

Cleanups:

* Update terminology in task names and erorr messages to be consitent with the parameter names: database password and Oracle metrics password (rather than SYS and agent passwords).
* Remove redundant password check at user creation time
* Expand the allowed special characters, though we still exclude `&` because of its special meaning in sqlplus.
* Update task names to include the role as per the regular standard

Testcases:
https://gist.github.com/mfielding/5e0c99e83bb7dfd6a94b457f45a7f5bb

### DIFF
--- a/roles/ora-host/tasks/validate_passwords.yml
+++ b/roles/ora-host/tasks/validate_passwords.yml
@@ -44,12 +44,12 @@
       # Must contain at least one lowercase letter
       - agent_pass_result.stdout is search('[a-z]')
       # Must contain at least one digit
-      - agent_pass_result.stdout is search('[0-8]')
+      - agent_pass_result.stdout is search('[0-9]')
       # Must contain at least one special character
       - agent_pass_result.stdout is search('[#$@%*_+=-]')
       # Must not contain common disallowed words (e.g., 'oracle') (No change needed here)
       - agent_pass_result.stdout | lower is not search('oracle')
-    fail_msg: "Oracle metrics user password failed Oracle complexity requirements. Check length (min 8), case mix, digits, and special characters."
+    fail_msg: "Oracle metrics user password failed Oracle complexity requirements. Check length (min 8), case mix, digits, and special characters #$@%*_+=-"
     success_msg: "Oracle metrics user password passed complexity validation."
   when:
     - not (agent_pass_result.skipped | default(false))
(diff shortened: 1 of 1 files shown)


roles/ora-host/tasks/validate_passwords.yml
```diff
@@ -35,16 +35,27 @@
...
+      - agent_pass_result.stdout is search('[A-Z]')
+      # Must contain at least one lowercase letter
+      - agent_pass_result.stdout is search('[a-z]')
+      # Must contain at least one digit
+      - agent_pass_result.stdout is search('[0-8]')
+      # Must contain at least one special character
+      - agent_pass_result.stdout is search('[#$@%*_+=-]')
+      # Must not contain common disallowed words (e.g., 'oracle') (No change needed here)
+      - agent_pass_result.stdout | lower is not search('oracle')
+    fail_msg: "Oracle metrics user password failed Oracle complexity requirements. Check length (min 8), case mix, digits, and special characters."
```
AlexBasinov, line 52: Can you please update the fail_msg to explicitly list the special characters (#$@%*_+=-) so the users don't have to check the Ansible code to understand why the validation failed?

roles/ora-host/tasks/validate_passwords.yml
```diff
@@ -35,16 +35,27 @@
...
+- name: validate_passwords | Validate Oracle metrics user password
+  ansible.builtin.assert:
+    that:
+      - agent_pass_result.stdout | length >= 8
+      # Must contain at least one uppercase letter
+      - agent_pass_result.stdout is search('[A-Z]')
+      # Must contain at least one lowercase letter
+      - agent_pass_result.stdout is search('[a-z]')
+      # Must contain at least one digit
+      - agent_pass_result.stdout is search('[0-8]')
```
AlexBasinov, line 47: This is likely a typo. We probably want to match any digit from 0 through 9

- agent_pass_result.stdout is search('[0-9]')

or alternatively, we can use the shorthand character class for digits: 

- agent_pass_result.stdout is search('\d')

mfielding: Done

AlexBasinov (approved):

PR #362, opened 2025-09-26
Add .tool-versions to declare runtime dependencies for SBOM

sahsagar-google:
This change adds a .tool-versions file to declare and pin the specific versions of runtime dependencies like Ansible and Python. This allows the build script to generate a more accurate SBOM by reading the required versions directly from this file.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,6 +1,9 @@
-# This file pins the required versions of runtime dependencies for the oracle-toolkit.
-# It is used by the build script to generate an accurate SBOM and ensures a
-# consistent environment for developers.
+# This file pins the required versions of runtime dependencies for the oracle-toolkit 
+# to ensure a consistent developer environment and to generate an accurate SBOM.
+#
+# The Python version specified is the minimum required by the pinned Ansible version, 
+# as per the Ansible support matrix:
+# https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix
 
-ansible 2.16.5
-python 3.9.2
+ansible 2.9
+python 3.5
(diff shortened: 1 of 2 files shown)


.tool-versions
```diff
@@ -0,0 +1,6 @@
+# This file pins the required versions of runtime dependencies for the oracle-toolkit.
+# It is used by the build script to generate an accurate SBOM and ensures a
+# consistent environment for developers.
+
+ansible 2.16.5
```
mfielding, line 5: The current minimum Ansible version is in https://github.com/google/oracle-toolkit/blob/5d6dfa766c090f1c869a23b4432a270cfcedacf3/group_vars/all.yml#L167

This can change from time to time though;  in the spirit of automatic SBOM scanning, is there any way to fetch it from that canonical location?

.tool-versions
```diff
@@ -0,0 +1,6 @@
...
+python 3.9.2
```
mfielding, line 6: oracle-toolkit doesn't have a specific Python version requirement, but rather whatever minimum Python version to run the minimum Ansible version.  The ansible-core support matrix is at https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix

group_vars/all.yml
```diff
@@ -164,7 +164,7 @@ proxy_setup: "{{ lookup('env','ORA_PROXY_SETUP') | lower | default('false',true)
 u01_lun: "{{ lookup('env','ORA_U01_LUN') | lower | default('',true) }}"
 
 #  Environment and infrastructure values (requirements) used for readiness checks
-minimum_ansible_version: 2.9
+minimum_ansible_version: "{{ lookup('file', playbook_dir + '/../.tool-versions') | regex_search('^ansible\\s+(.*)$', '\\1') | first }}"
```
mfielding, line 167: I think I've changed my mind here:  we don't change ansible minima often, and it looks like the file logic might not always work, so can we just go back to hard-coding to 2.9?  Thanks!

mfielding (approved):

PR #430, opened 2026-05-20
Implement storage pools options for Oracle deployments in terraform templates

DanielUlisses:
### Change Description:
Provide with terraform to create and use storage pools within the code.
Provide an existing storage pool and use the storage pool for deployment.

### Solution  Overview:
Implements storage pool support on the toolkit terraform configuration

supports 'hyperdisk-balanced' or 'hyperdisk-throughput' storage pools

### Test Commands:
```
terraform init
terraform apply -var-file=../presubmit_tests/storage-pool.tfvars
```
Test Prep:
The following variable object must be provided to the terraform

```
storage_pool = {
  enabled                                     = true
  storage_pool_type                     = "hyperdisk-balanced"
  deletion_protection                    = false
  capacity_provisioning_type        = "ADVANCED"
  performance_provisioning_type = "ADVANCED"
  pool_provisioned_capacity_gb   = 10240 # minimum 10 TiB
  pool_provisioned_iops                = 10000
  pool_provisioned_throughput      = 1024
}
```
An example file was added at  presubmit_tests/storage-pool.tfvars

### Test 1:  details:
Terraform will create the oracle servers with all data disks pointing to the storage pools

The implementation should be transparent for other components of toolkit

### Expected Results:
When enable terraform outputs will change adding storage pools references

Outputs:

```
control_node_log_url = "https://console.cloud.google.com/logs/query;query=resource.labels.instance_id%3D4410714249636111124;duration=P30D?project=oval-replica-444520-i8"
database_vm_names = [
  "gto59ora-8-1",
  "gto59ora-8-2",
]
storage_pool_zone1_self_link = "projects/oval-replica-444520-i8/zones/us-east1-b/storagePools/gto59ora-8-pool-z1"
storage_pool_zone2_self_link = "projects/oval-replica-444520-i8/zones/us-east1-c/storagePools/gto59ora-8-pool-z2"
```


mfielding:
@DanielUlisses please sign the Google CLA: https://cla.developers.google.com/

mfielding:
The presubmit failures are at least partially due to the underlying cloud service account missing privileges to manipulate storage pools.  I'm working on an internal change to give minimal privileges via a custom role.

presubmit_tests/storage-pools-on-gcp.sh
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+source presubmit_tests/infra-manager-lib.sh || {
+  echo "Error: cannot source common library" >&2
+  exit 1
+}
+
+instance_name="github-presubmit-stpools-${BUILD_ID}"
+deployment_name="presubmit-stpools-${BUILD_ID}"
+tfvars_file="./presubmit_tests/storage-pools.tfvars"
```
mfielding: There is a typo here referencing a plural `storage-pools.tfvars` file, which causes the presubmit run to fail immediately as the file in the repo is singular (`storage-pool.tfvars`).

```suggestion
tfvars_file="./presubmit_tests/storage-pool.tfvars"

File: presubmit_tests/storage-pool.tfvars [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+enable_ar_repo   = true
+
+storage_pool = {
+  enabled                       = true
+  storage_pool_type             = "hyperdisk-balanced"
+  capacity_provisioning_type    = "ADVANCED"
+  performance_provisioning_type = "ADVANCED"
+  pool_provisioned_capacity_gb  = 10240 # minimum 10 TiB
+  pool_provisioned_iops         = 10000
+  pool_provisioned_throughput   = 1000
```
mfielding: The GCE API rejects `1000` MB/s because provisioned throughput for Hyperdisk Balanced storage pools must be a multiple of `1024` (e.g., `1024` MB/s = 1 GB/s). This causes a 400 BadRequest error during `terraform apply`.

We should also set `deletion_protection = false` here so the CI/CD runner can clean up cleanly if the test fails.

```suggestion
  deletion_protection           = false
  pool_provisioned_throughput   = 1024

File: terraform/variables.tf [hunk 3]
```diff
@@ -237,6 +237,36 @@ variable "reco_disk" {
...
+  }
+
+  validation {
+    condition = var.storage_pool == null || (
+      try(var.storage_pool.pool_provisioned_capacity_gb, 10240) >= 10240
+    )
+    error_message = "storage_pool.pool_provisioned_capacity_gb must be at least 10240 (10 TiB minimum)."
+  }
+}
+
```
mfielding, line 291: ### Part 1: Restructure Variables
Since this is a brand-new feature, let's restructure the variables to distinguish between auto-creating a pool (for sandboxes) and using existing shared pools (for production). Please replace the `storage_pool` variable block with these two:

```hcl
# 1. For auto-creating a pool (Sandbox/Testing/CI)
variable "create_storage_pool" {
  description = "Configuration for auto-creating a storage pool for sandbox/testing. If enabled, a new pool will be created in each deployment zone."
  type = object({
    enabled                       = optional(bool, false)
    storage_pool_type             = optional(string, "hyperdisk-balanced")
    capacity_provisioning_type    = optional(string, "advanced")
    deletion_protection           = optional(bool, true)
    performance_provisioning_type = optional(string, "advanced")
    pool_provisioned_capacity_gb  = optional(number, 10240)
    pool_provisioned_iops         = optional(number, 10000)
    pool_provisioned_throughput   = optional(number, 1024)
  })
  default = null
}

# 2. For using pre-existing shared pools (Production)
variable "existing_storage_pools" {
  description = "Map of existing storage pools to use, keyed by zone (e.g. {'us-central1-b' = 'projects/.../storagePools/pool-b'}). If provided, disks in these zones will be attached to the specified pools."
  type        = map(string)
  default     = {}
}
```

---

### Part 2: Update Examples (Action Required)
Please also modify the existing example files **`terraform/terraform.tfvars.xfs.example`** and **`terraform/terraform.tfvars.asm.example`** (under the *Disk Settings* section) to add this commented-out block, demonstrating how users can attach their disks to pre-existing shared storage pools:

```tfvars
# --- Optional Shared Storage Pools ---
# To attach the database disks to pre-existing shared Hyperdisk Storage Pools (recommended for production):
# existing_storage_pools = {
#   "us-central1-b" = "projects/my-project/zones/us-central1-b/storagePools/shared-pool-z1"
#   "us-central1-c" = "projects/my-project/zones/us-central1-c/storagePools/shared-pool-z2"
# }
```

terraform/main.tf
```diff
@@ -112,6 +112,12 @@ locals {
   # Concatenetes both lists to be passed down to the instance module
   additional_disks = concat(local.fs_disks, local.asm_disks)
 
+  # Storage pool helpers
+  storage_pool_enabled = var.storage_pool != null && try(var.storage_pool.enabled, false)
```
mfielding: Let's update the local helpers to drive the conditional creation logic:

```hcl
  # Storage pool enabled if either auto-create is requested OR existing pools are mapped
  storage_pool_enabled = (var.create_storage_pool != null && try(var.create_storage_pool.enabled, false)) || length(var.existing_storage_pools) > 0
  
  # Only create new pools if create is enabled AND no existing pools are mapped
  create_pool_enabled  = var.create_storage_pool != null && try(var.create_storage_pool.enabled, false) && length(var.existing_storage_pools) == 0

File: terraform/main.tf [hunk 5]
```diff
@@ -254,6 +278,65 @@ resource "google_compute_instance_from_template" "database_vm" {
   }
 }
 
+# -----------------------------------------------------------------------------
+# Hyperdisk Storage Pool (optional)
+# -----------------------------------------------------------------------------
+
+resource "google_compute_storage_pool" "zone1" {
```
mfielding, line 288: Update the storage pool resources to only create if `create_pool_enabled` is true:

```hcl
resource "google_compute_storage_pool" "zone1" {
  count   = local.create_pool_enabled ? 1 : 0
  project = var.project_id
  name    = "${var.instance_name}-pool-z1"
  zone    = var.zone1

  storage_pool_type             = "projects/${var.project_id}/zones/${var.zone1}/storagePoolTypes/${var.create_storage_pool.storage_pool_type}"
  capacity_provisioning_type    = var.create_storage_pool.capacity_provisioning_type
  deletion_protection           = var.create_storage_pool.deletion_protection
  performance_provisioning_type = var.create_storage_pool.performance_provisioning_type
  pool_provisioned_capacity_gb  = var.create_storage_pool.pool_provisioned_capacity_gb
  pool_provisioned_iops         = var.create_storage_pool.storage_pool_type == "hyperdisk-balanced" ? var.create_storage_pool.pool_provisioned_iops : null
  pool_provisioned_throughput   = var.create_storage_pool.pool_provisioned_throughput
}

resource "google_compute_storage_pool" "zone2" {
  count   = (local.create_pool_enabled && local.is_multi_instance) ? 1 : 0
  project = var.project_id
  name    = "${var.instance_name}-pool-z2"
  zone    = var.zone2

  storage_pool_type             = "projects/${var.project_id}/zones/${var.zone2}/storagePoolTypes/${var.create_storage_pool.storage_pool_type}"
  capacity_provisioning_type    = var.create_storage_pool.capacity_provisioning_type
  deletion_protection           = var.create_storage_pool.deletion_protection
  performance_provisioning_type = var.create_storage_pool.performance_provisioning_type
  pool_provisioned_capacity_gb  = var.create_storage_pool.pool_provisioned_capacity_gb
  pool_provisioned_iops         = var.create_storage_pool.storage_pool_type == "hyperdisk-balanced" ? var.create_storage_pool.pool_provisioned_iops : null
  pool_provisioned_throughput   = var.create_storage_pool.pool_provisioned_throughput
}

File: terraform/main.tf [hunk 6]
```diff
@@ -254,6 +278,65 @@ resource "google_compute_instance_from_template" "database_vm" {
...
+resource "google_compute_disk" "pool_disks" {
+  for_each = local.pool_disk_instances
+  project  = var.project_id
+  name     = each.key
+  zone     = each.value.zone
+  type     = each.value.disk_type
+  size     = each.value.disk_size_gb
+  labels   = each.value.disk_labels
+
+  storage_pool = (
```
mfielding, line 327: Update the disk attachment logic to dynamically select between the pre-existing pool map or the newly created pool:

```hcl
  storage_pool = (
    length(var.existing_storage_pools) > 0
    ? lookup(var.existing_storage_pools, each.value.zone, null)
    : (
        each.value.zone == var.zone1
        ? one(google_compute_storage_pool.zone1[*].id)
        : one(google_compute_storage_pool.zone2[*].id)
      )
  )

File: docs/terraform.md
Comment (mfielding) [inline docs/terraform.md:72]: Please add a dedicated section documenting the Storage Pools feature, as it is a pure infrastructure/Terraform-level configuration:

```markdown
## Hyperdisk Storage Pools (Optional)

Hyperdisk Storage Pools allow you to aggregate storage capacity and performance across multiple virtual machines. This is highly recommended for multi-database environments to improve storage efficiency and reduce costs.

The toolkit supports two modes of operation:

1.  **Auto-Create (Sandbox / Testing)**: The toolkit creates a dedicated storage pool on the fly for the deployment. Configure this using the `create_storage_pool` block in your `.tfvars`. Note that GCE storage pools have a 10 TiB minimum capacity constraint and are billed for provisioned capacity.
2.  **Existing Shared Pool (Production / Shared Environments)**: Recommended for production. The toolkit attaches the database disks directly to pre-created shared storage pools, bypassing the need for the deployment service account to have creation permissions. Configure this by mapping your zones to existing pool IDs in the `existing_storage_pools` map.

Example of using existing pools in `terraform.tfvars`:
\`\`\`tfvars
existing_storage_pools = {
  "us-central1-b" = "projects/my-project/zones/us-central1-b/storagePools/shared-pool-z1"
  "us-central1-c" = "projects/my-project/zones/us-central1-c/storagePools/shared-pool-z2"
}
\`\`\`

**Review by mfielding — COMMENTED** [review]
Hi @DanielUlisses,

Thank you for this excellent contribution! Hyperdisk Storage Pools are a great addition to the Oracle Toolkit.

I have successfully validated the core infrastructure provisioning and Oracle database installation using your Terraform template. The infrastructure deploys correctly, and the database installs successfully once a few minor API and scripting bugs are resolved (detailed in the line comments).

I also verified that the provider upgrade to `~> 6.31.0` in `versions.tf` is correct and necessary, as the `google_compute_storage_pool` resource is not available in older versions.

However, I have some critical architectural and documentation feedback that needs to be addressed before we can merge this:

### 1. Architectural Feedback: Support for "Existing" Storage Pools (Recommended)
Currently, the PR always creates a new storage pool dynamically per deployment. While this works for isolated sandboxes, it is an anti-pattern for production:
*   **Cost & Efficiency**: Storage pools have a **10 TiB minimum capacity**. Creating a new pool per database deployment is highly inefficient and expensive.
*   **Production Pattern**: In production, a platform team pre-creates a shared storage pool, and multiple independent deployments attach their disks to this existing pool.

**Proposal**: Since this is a brand-new feature, we don't have to worry about backward compatibility. I suggest we restructure the parameters to support both use cases cleanly:
1.  **`create_storage_pool`**: A parameter explicitly designed for auto-creating a pool on-the-fly (Sandbox/Testing mode).
2.  **`existing_storage_pools`**: A map of `zone => pool_id` allowing the toolkit to attach database disks directly to pre-existing, shared storage pools (Production mode).
*(See code suggestions in the comments on `variables.tf` and `main.tf`).*

### 2. Presubmit Test Optimization & Cleanup
Running a dedicated, parallel presubmit test (`storage-pools-on-gcp.sh`) that boots a new VM and installs Oracle takes ~45 minutes. 
Instead of a separate script, we should **incorporate storage pool testing into the existing `free-edition-on-gcp.sh` test** by simply enabling `create_storage_pool` in `free-edition.tfvars` (with deletion protection disabled for clean CI/CD cleanup). This validates the pool creation and attachment logic without the overhead of an extra parallel run.

Consequently, we should **delete the separate `storage-pools-on-gcp.sh` and `storage-pool.tfvars` files** to keep the test suite clean.

### 3. Documentation & Examples
*   **Examples**: Instead of adding a new standalone tfvars example (which isn't usable on its own), please add a commented-out `existing_storage_pools` block in the existing `terraform/terraform.tfvars.xfs.example` and `terraform/terraform.tfvars.asm.example` files.
*   **Documentation**: Please document this feature in `docs/terraform.md` (since it is a pure infrastructure/Terraform feature) rather than `docs/user-guide.md`.

Please see the specific line comments below for the bug fixes, code suggestions, and documentation templates.

**Comment by DanielUlisses** [comment]
Thanks @mfielding  for your time and valuable comments and suggestion in the review
I liked to suggestions to allow both pre-created and new storage pools as a parallel option without need to import resources
I will look up on them and implement the changes

File: presubmit_tests/storage-pools-on-gcp.sh (hunk 1 above)
Comment (DanielUlisses) [reply presubmit_tests/storage-pools-on-gcp.sh]: it should be fixed on the new commit

File: presubmit_tests/storage-pool.tfvars (hunk 2 above)
Comment (DanielUlisses) [reply presubmit_tests/storage-pool.tfvars]: it should be fixed on the new commit

File: terraform/variables.tf (hunk 3 above)
Comment (DanielUlisses) [reply terraform/variables.tf:291]: it should be fixed on the new commit

File: terraform/main.tf (hunk 4 above)
Comment (DanielUlisses) [reply terraform/main.tf]: it should be fixed on the new commit

File: terraform/main.tf (hunk 5 above)
Comment (DanielUlisses) [reply terraform/main.tf:288]: it should be fixed on the new commit

File: terraform/main.tf (hunk 6 above)
Comment (DanielUlisses) [reply terraform/main.tf:327]: it should be fixed on the new commit

File: docs/terraform.md
Comment (DanielUlisses) [reply docs/terraform.md:72]: terraform documentation section was added

**Comment by DanielUlisses** [comment]
To make sure we don't get false positives in the tests I rebased it from current master and pushed it again

File: docs/terraform.md [hunk 7]
```diff
@@ -16,6 +16,7 @@ This setup supports the deployment of the following configurations:
 - Network configuration using specified subnets and NAT IPs
 - Custom startup scripts for VM initialization (Google Cloud metadata [startup scripts](https://cloud.google.com/compute/docs/instances/startup-scripts/linux))
 - Ansible automation for post-provisioning configuration tasks
+- Storage Pools options for VM deployments
```
mfielding: nit: I'd remove this line, as storage pools are mentioned later in this doc anyway.

presubmit_tests/free-edition.tfvars
```diff
@@ -53,3 +53,16 @@ ora_db_name = "orcl"
 ora_db_domain = "test1.example_domain01.com"
 ora_db_container = true
 
+
+# --- Optional Shared Storage Pools ---
```
mfielding: nit: in the context of a presubmit test, it isn't "optional"; as it's the configuration we're using.

terraform/main.tf
```diff
@@ -354,6 +444,17 @@ resource "google_compute_instance" "control_node" {
       )
       error_message = "The 'enable_ar_repo' feature is enabled, but Private Google Access (PGA) is not enabled on the target subnetwork(s). PGA is required for internal access to Artifact Registry."
     }
+
```
mfielding, line 447: There is a logical bug here that will block users who want to use pre-existing storage pools (Production Mode). 

If `existing_storage_pools` is used, `create_storage_pool.enabled` is `false`, which causes this precondition to always evaluate to `false` and abort the deployment. We should only perform this static disk-type check if we are actually *creating* the pool in this template.

Please change `!local.storage_pool_enabled` to `!local.create_pool_enabled`:

```suggestion
    precondition {
      condition = !local.create_pool_enabled || (
        var.oracle_home_disk.type == var.create_storage_pool.storage_pool_type &&
        var.data_disk.type == var.create_storage_pool.storage_pool_type &&
        var.reco_disk.type == var.create_storage_pool.storage_pool_type &&
        var.swap_disk_type == var.create_storage_pool.storage_pool_type
      )
      error_message = "When storage_pool is enabled, oracle_home_disk.type, data_disk.type, reco_disk.type, and swap_disk_type must all match storage_pool.storage_pool_type."
    }
```

DanielUlisses: makes sense. 
this was the section that was causing the null when we are not using storage pools, let me adjusted that. 
good catch, thanks

mfielding: Hi @DanielUlisses,

I see that in commit fe75eb2 you added a default non-null object for `create_storage_pool` and added `var.create_storage_pool.enabled` to the precondition to avoid null-pointer errors. 

While this prevents Terraform syntax errors, it unfortunately **does not fix the logical bug** for the production use case. 

### Why it still fails:
If a user wants to use pre-existing storage pools (Production Mode), they will define `existing_storage_pools` and leave `create_storage_pool.enabled` as `false` (default). In this scenario:
1.  `local.storage_pool_enabled` evaluates to `true` (since `existing_storage_pools` is populated).
2.  Therefore, `!local.storage_pool_enabled` evaluates to `false`.
3.  Terraform then evaluates the right side of the `||`: `var.create_storage_pool.enabled && ...`
4.  Since `var.create_storage_pool.enabled` is `false`, the right side evaluates to `false`.
5.  The final condition evaluates to `false || false` = `false`, causing the deployment to **fail the precondition and abort**.

### The Fix:
We should only validate the disk types against the storage pool type if the template is actually responsible for *creating* the pool. If we are using existing pools, we cannot validate this statically.

Please change `!local.storage_pool_enabled` to **`!local.create_pool_enabled`** in the precondition. This resolves the bug for existing pools while still enforcing the check when creating a new pool:

```suggestion
    precondition {
      condition = !local.create_pool_enabled || (
        var.oracle_home_disk.type == var.create_storage_pool.storage_pool_type &&
        var.data_disk.type == var.create_storage_pool.storage_pool_type &&
        var.reco_disk.type == var.create_storage_pool.storage_pool_type &&
        var.swap_disk_type == var.create_storage_pool.storage_pool_type
      )
      error_message = "When storage_pool is enabled, oracle_home_disk.type, data_disk.type, reco_disk.type, and swap_disk_type must all match storage_pool.storage_pool_type."
    }
```

DanielUlisses: this change is now applied in #2f1b3c9 and #f4a8dc11, let's wait for the tests but seems correct on my local validations

.claude/agents/README.md
```diff
@@ -0,0 +1,41 @@
+# Local Agents — Override AgentSpec
```
mfielding: I don't think you intended to include this in the PR :-)

DanielUlisses: Yes, I don't let me remove it from here

mfielding: FWIW

To prevent local AI assistant metadata (like `.claude/` or `.gemini/`) from accidentally being staged in the future, we recommend adding these patterns to your **global gitignore** file rather than the project's shared `.gitignore`.
You can set this up locally by running:
```bash
# Create a global gitignore file if you don't have one
mkdir -p ~/.config/git
touch ~/.config/git/ignore
# Add the patterns you want to ignore globally
echo ".claude/" >> ~/.config/git/ignore
echo ".gemini/" >> ~/.config/git/ignore
echo ".cursor/" >> ~/.config/git/ignore
# Configure git to use this file globally
git config --global core.excludesfile ~/.config/git/ignore

mfielding (approved):

PR #432, opened 2026-05-26
fix(presubmit): resolve gcloud logging tail crash loop

mfielding:
Forces the Google Cloud CLI launcher to run on the prepackaged Python 3.14 environment (where grpcio is preinstalled) rather than defaulting to the system Python 3.12 environment (which is pulled in dynamically on Alpine by apk py3-pip package, but lacks grpcio).

This should fix loops like https://oss.gprow.dev/view/gs/gcp-oracle-prow-bucket/pr-logs/pull/google_oracle-toolkit/431/oracle-toolkit-install-26ai-data-guard-on-gcp/2059295359544332288

```
2026-05-26 15:29:55    Starting gcloud logging tail session...
ERROR: (gcloud.alpha.logging.tail) Please ensure that the gRPC module is installed and the environment is correctly configured. Run:
  sudo pip3 install grpcio
and set:
  export CLOUDSDK_PYTHON_SITEPACKAGES=1
For more information, see https://cloud.google.com/logging/docs/reference/tools/gcloud-logging#install_live_tailing
```

Also escapes dynamic date evaluations inside the Prow logging heredoc process group, ensuring live logs stream correctly in real-time.

### DIFF
--- a/presubmit_tests/infra-manager-lib.sh
+++ b/presubmit_tests/infra-manager-lib.sh
@@ -25,11 +25,10 @@ setup_vars() {
     echo "\$deployment_name is missing." >&2
     exit 1
   fi
-  apk add --no-cache zip curl py3-pip expect || exit 1
+  apk add --no-cache zip curl expect || exit 1
 
   # Force gcloud to use the preinstalled Python 3.14 environment
-  # which already has grpcio installed, preventing version mismatch conflicts
-  # with the system Python 3.12 dynamically pulled in by apk for py3-pip.
+  # which already has grpcio prepackaged, ensuring robust execution.
   export CLOUDSDK_PYTHON=python3
   export CLOUDSDK_PYTHON_SITEPACKAGES=1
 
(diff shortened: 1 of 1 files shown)


presubmit_tests/infra-manager-lib.sh
```diff
@@ -26,6 +26,13 @@ setup_vars() {
     exit 1
   fi
   apk add --no-cache zip curl py3-pip expect || exit 1
+
+  # Force gcloud to use the preinstalled Python 3.14 environment
+  # which already has grpcio installed, preventing version mismatch conflicts
```
AlexBasinov: Does this mean that we no longer need to install grpcio with pip? if so, should we remove the following command from line #129? 

pip3 install grpcio --break-system-packages || exit 1

mfielding: Nice catch.  Actually, the news gets better.  If we don't need pip3, we don't need to install our own Python at all.  The built-in Python 3.14 has everything we need.  And with a single version of Python, we don't need the gcloud overrides either.

AlexBasinov (approved):

PR #436, opened 2026-06-11
Add ora_db_unique_name Terraform variable to set DB_UNIQUE_NAME

GMafra:
### Change Description:
Add terraform support for setting the Oracle database unique name (DB_UNIQUE_NAME).
Pass the new `--ora-db-unique-name` flag from terraform through to `install-oracle.sh`.

This brings the terraform layer to parity with `install-oracle.sh`, which already
supports `--ora-db-unique-name` for both Primary and Standby databases.

### Solution Overview:
Implements `ora_db_unique_name` support on the toolkit terraform configuration

Optional variable; works for both Primary and Standby databases. When left empty the
existing ansible default applies (`db_name` for Primary, `db_name_s` for Standby)

Test Commands:
```
terraform init
terraform apply -var-file=terraform.tfvars
```
Test Prep:
The following variable can be provided to terraform (optional)

```
ora_db_unique_name = "ora19p"
```
An example value was added to the `terraform.tfvars.*.example` files

### Test 1: details:
Terraform renders the `install-oracle.sh` command with `--ora-db-unique-name` when the
variable is set, and omits it when left empty.

The implementation should be transparent for other components of toolkit

### Expected Results:
When `ora_db_unique_name` is set, the generated `install-oracle.sh` command includes
`--ora-db-unique-name <value>`, and the database comes up with the expected
`DB_UNIQUE_NAME` for both Primary and Standby installs.

[oratk59-17 terraform test results](https://gist.github.com/GMafra/0514cdadeab4f1d253de65b58c716105)

[database values](https://gist.github.com/GMafra/98780c745fbb61a78ab1b46220c71d2e)


terraform/terraform.tfvars.asm.example
```diff
@@ -69,6 +69,7 @@ ora_release                = "latest"
 ora_edition                = "EE"
 ora_backup_dest            = "+RECO"
 ora_db_name                = "orcl"
+ora_db_unique_name         = "ora19p"
```
mfielding: nit: in a newly-created, single-instance database, we'd typically set ora_db_name and ora_db_unique_name to the same value.

mfielding (commented):
Looks good; just on nit.

terraform/terraform.tfvars.asm.example
GMafra: @mfielding updated

GMafra:
@mfielding nit addressed

mfielding (approved):

PR #439, opened 2026-06-25
Add GCNV iSCSI storage backend

kanikaj:
## Summary

Adds Google Cloud NetApp Volumes (GCNV) iSCSI as an additional storage backend
alongside Hyperdisk for the Oracle DATA/RECO disk groups, selected with
`storage_backend = "gcnv"`. The default stays `hyperdisk` and the existing
Hyperdisk path is unchanged — all GCNV resources are gated on `local.is_gcnv`,
so current deployments are unaffected.

## Why a two-step flow

An iSCSI host group must authorize a host's real initiator IQN, which only
exists after the VM boots. To keep provisioning fully automated, it is split
into two steps:

- **Step 1** (`provision-gcnv.sh` → `terraform apply -var gcnv_provision_only=true`):
  creates PSA + the NetApp Flex pool + iSCSI LUNs (no host group yet) + the DB
  VMs + a short-lived provisioning control node. That node, from inside the VPC,
  SSHes each DB VM, reads its real IQN, creates the per-node host group, attaches
  the LUNs via the NetApp API, then self-deletes.
- **Step 2** is the existing, unmodified install flow. The LUNs are already
  authorized, so the new `iscsi-multipath` role just logs in and presents
  `/dev/mapper/*` devices to ASM.

## What's included

- `terraform/modules/gcnv_psa` — PSA range + service-networking connection
  (skippable via `gcnv_create_psa=false` when PSA already exists on the VPC).
- `terraform/modules/netapp_iscsi` — Flex pool + iSCSI LUNs.
- Provisioning control node + startup script (`terraform/scripts/gcnv-provision.sh.tpl`)
  for IQN discovery, host-group creation, and LUN attachment.
- `roles/iscsi-multipath` Ansible role (no-ops when no portals are passed, so
  Hyperdisk runs are untouched).
- `provision-gcnv.sh` wrapper and additive `install-oracle.sh` flags
  (`--iscsi-portals`, `--gcnv-host-map-json`).
- `docs/gcnv-storage-backend.md` — prerequisites, two-step usage, validation,
  failure/recovery, and known limitations.
- A `moved` block so existing users don't see `control_node` recreated on upgrade.

## Backward compatibility

- Defaults: `storage_backend = "hyperdisk"`, `gcnv_provision_only = false`.
- No changes to the Hyperdisk Terraform path or the Step 2 install logic beyond
  additive, opt-in flags/role that no-op without GCNV inputs.
- Provider floor raised to `>= 7.0.0` for the GCNV resources.

[...]

PR #441, opened 2026-06-30
feat(tools): allow GIM update in-place and improve revision monitoring

mfielding:
Based on testing for #439, this PR introduces improvements to the `deploy-infra-manager.sh` script to support iterative development:

1. **In-place Updates**: Allows updating an existing deployment in-place instead of failing immediately, enabling faster iteration.
2. **Silenced Progress**: Silences gcloud progress indicators in logs for cleaner output.
3. **Improved Revision Monitoring**: Robustly detects and monitors new revisions (recovering from failures and using fallback listing if needed).
4. **Provider Source**: Adds support for specifying a custom `--provider-source`.

### DIFF
--- a/tools/deploy-infra-manager.sh
+++ b/tools/deploy-infra-manager.sh
@@ -18,6 +18,9 @@
 
 set -e
 
+# Suppress gcloud progress indicators and spinners in logs
+export CLOUDSDK_CORE_INTERACTIVE_UX_STYLE=OFF
+
 # --- Path Calculations ---
 SCRIPT_PATH=$(readlink -f "$0")
 SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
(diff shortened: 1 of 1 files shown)


AlexBasinov (approved):

AlexBasinov:
Just out of curiosity, what prompted you to expose the `--provider-source` flag? Did you run into any issues when using the default provider?

mfielding:
> Just out of curiosity, what prompted you to expose the `--provider-source` flag? Did you run into any issues when using the default provider?

Yes, the addition of the `--provider-source` flag was prompted by network environment constraints.

### The Problem:
In many restricted or private Google Cloud projects (including certain strict sandbox/testing projects and corporate workloads), external egress to the public internet is disabled or highly restricted. 

When Infrastructure Manager runs its deployment pipeline (which does a `terraform init` under the hood), it tries to fetch the required Terraform providers (like `hashicorp/google`, `hashicorp/null`, `hashicorp/random`) from the public **HashiCorp Terraform Registry** (`registry.terraform.io`). If egress is blocked, GIM fails immediately at the initialization stage because it cannot download the providers.

### The Solution:
Exposing `--provider-source` allows us to pass this flag through to the underlying `gcloud infra-manager deployments apply` command.

By specifying `--provider-source=SERVICE_MAINTAINED`, we tell Infrastructure Manager to bypass the public registry and instead use the **Google Cloud-maintained/pre-cached provider mirrors** hosted internally within Google's network. 

I exposed this flag in the script to ensure that the deployment tool remains flexible enough to work seamlessly in both public-egress-enabled sandbox environments and strict private projects where public registry lookups are blocked.

PR #448, opened 2026-08-26
FSFO observer provisioning and configuration

pythianakhmadeev:
## Change Description:
Provide an option to configure a separate machine as an Observer for a Fast-Start Failover configuration of the Data Guard setup.

## Solution Overview:
Add a new parameter `--config-observer` to setup and configure an FSFO Observer on a dedicated host.

## Test Commands:

Oracle 19c and 26ai Data Guard on FS and ASM as following (example for 19c):
```
export INSTANCE_IP_ADDR=
export STANDBY_IP_ADDR=
export FSFO_IP_ADDR=

# Data Guard provisioning
./install-oracle.sh \
  --ora-edition EE \
  --ora-version 19 \
  --ora-disk-mgmt FS \
  --ora-db-name ora19 --ora-db-unique-name ora19p \
  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"},{"purpose":"data","blk_device":"/dev/disk/by-id/google-oracle-data-1","name":"u03","fstype":"xfs","mount_point":"/u03","mount_opts":"nofail"},{"purpose":"reco","blk_device":"/dev/disk/by-id/google-oracle-reco-1","name":"u04","fstype":"xfs","mount_point":"/u04","mount_opts":"nofail"}]' \
  --ora-data-destination "/u03/app/oracle/oradata" \
  --ora-reco-destination "/u04/app/oracle/fast_recovery_area" \
  --backup-dest "/u04/app/oracle/backups"

./install-oracle.sh \
  --instance-ip-addr ${STANDBY_IP_ADDR} \
  --primary-ip-addr ${INSTANCE_IP_ADDR} \
  --cluster-type DG \
  --ora-edition EE --ora-version 19 \
  --ora-disk-mgmt FS \
  --ora-db-name ora19 --ora-db-unique-name ora19s1 \
  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"},{"purpose":"data","blk_device":"/dev/disk/by-id/google-oracle-data-1","name":"u03","fstype":"xfs","mount_point":"/u03","mount_opts":"nofail"},{"purpose":"reco","blk_device":"/dev/disk/by-id/google-oracle-reco-1","name":"u04","fstype":"xfs","mount_point":"/u04","mount_opts":"nofail"}]' \
  --ora-data-destination "/u03/app/oracle/oradata" \
  --ora-reco-destination "/u04/app/oracle/fast_recovery_area" \
  --backup-dest "/u04/app/oracle/backups"

# FSFO installation and configuration
./install-oracle.sh \
  --config-observer \
  --instance-ip-addr ${FSFO_IP_ADDR} \
  --primary-ip-addr ${INSTANCE_IP_ADDR} \
  --cluster-type DG \
  --ora-edition CLIENT --ora-version 19 \
  --ora-disk-mgmt FS \
  --ora-db-name ora19 \
  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"}]'
```

## Expected Result:
An FSFO Data Guard configuration is successfully deployed


roles/dg-observer/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+  block:
+    - name: Copy observer start script
+      template:
+        src: "{{ observer_script }}.j2"
+        dest: "{{observer_script_dir}}/{{ observer_script }}"
+        owner: "{{ oracle_user }}"
+        group: "{{ oracle_group }}"
+        mode: "ug=rwx,o="
+
+    - name: Start the observer
```
mfielding: Rather than running manually, can you configure it as a systemd service, so that it runs on startup?

docs/disaster-recovery-configurations.md
```diff
@@ -47,6 +49,43 @@ export STANDBY_IP_ADDR=10.0.10.102
...
+  --prep-host \
+  --instance-ip-addr ${STANDBY_IP_ADDR} \
+  --instance-hostname standby-server-19c \
+  --ora-version 19 \
+  --ora-disk-mgmgt FS \
+  --ora-swlib-bucket gs://[BUCKET_NAME] \
+  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"}]'  
+
+./install-oracle.sh \
+  --install-sw \
```
mfielding: For a FSFO node, we should install the Oracle client rather than a full database server install.  BOth to save time/disk space and to make the licensing boundary clearer.

docs/disaster-recovery-configurations.md
```diff
@@ -47,6 +49,43 @@ export STANDBY_IP_ADDR=10.0.10.102
...
+  --instance-ip-addr ${STANDBY_IP_ADDR} \
+  --instance-hostname standby-server-19c \
+  --ora-version 19 \
+  --ora-disk-mgmgt FS \
+  --ora-swlib-bucket gs://[BUCKET_NAME] \
+  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"}]' 
+  
+./install-oracle.sh \
+  --config-observer \
```
mfielding, line 60: Given that a FSFO node shouldn't be on a standby at all, but rather a third location near the real database clients, should we consider a single `--config-observer` or `--install-observer` that does the full client and dependency install?

docs/disaster-recovery-configurations.md
```diff
@@ -47,6 +49,43 @@ export STANDBY_IP_ADDR=10.0.10.102
...
 
+<a name="create-data-guard-standby"></a>Create the Data Guard Observer:
+
+```bash
+# this is the Observer IP address
+export STANDBY_IP_ADDR=10.0.10.103
+export PRIMARY_IP_ADDRESS=10.0.10.101
+
+./install-oracle.sh \
+  --prep-host \
```
mfielding: I think we want a much more lightweight install for a FSFO observer than for a database host.  For example, no hugepages, probably no disk prep, etc.

docs/disaster-recovery-configurations.md
```diff
@@ -47,6 +49,43 @@ export STANDBY_IP_ADDR=10.0.10.102
...
+  --instance-ip-addr ${STANDBY_IP_ADDR} \
+  --instance-hostname standby-server-19c \
+  --ora-version 19 \
+  --ora-disk-mgmgt FS \
+  --ora-swlib-bucket gs://[BUCKET_NAME] \
+  --ora-data-mounts-json '[{"purpose":"software","blk_device":"/dev/disk/by-id/google-oracle-disk-1","name":"u01","fstype":"xfs","mount_point":"/u01","mount_opts":"nofail"}]'
```
mfielding: For simplicity, we may want to avoid a separate software mount at all, given all we need to install is an Oracle client; can it just sit on the OS root mount?

roles/dg-observer/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+      set_fact:
+        dg_wallet_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
+
+    - name: Create a wallet
+      shell: |
+        set -o pipefail
+        rm -rf "{{ wallet_dir }}"/*
+        echo -e "${WALLET_PWD}\n${WALLET_PWD}\n" | mkstore -wrl "{{ wallet_dir }}" -create
+        for x in {{ db_names.stdout }}; do
+          echo -e "${OBSERVER_PWD}\n${OBSERVER_PWD}\n${WALLET_PWD}\n" |
```
mfielding: Thie echo command puts the password on the command line, expositing it to any other user in the system (while it's running) via /proc/cmdline.  Can we secure it better?

roles/dg-observer/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+        echo -e "${WALLET_PWD}\n" | orapki wallet create -wallet "{{ wallet_dir }}" -auto_login_local
+        chmod 600 "{{ wallet_dir }}"/*
+      environment:
+        ORACLE_HOME: "{{ oracle_home }}"
+        ORACLE_SID: "{{ oracle_sid }}"
+        PATH: "{{ oracle_home }}/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin"
+        WALLET_PWD: "{{ dg_wallet_pass }}"
+        OBSERVER_PWD: "{{ dg_observer_pass }}"
+
+    - name: Store wallet password in a text file
```
mfielding: Do we even need to save the wallet password at all?  I think it should be possible to use the wallet without the password.  And for future changes, the wallet could be regenerated.

roles/dg-observer/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+        mode: '0700'
+
+    - name: Generate random password for wallet
+      set_fact:
+        dg_wallet_pass: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits') }}0#_"
+
+    - name: Create a wallet
+      shell: |
+        set -o pipefail
+        rm -rf "{{ wallet_dir }}"/*
```
mfielding: Is this safe?  Would we be wiping any existing credentials here?

roles/dg-observer/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+        src: "{{ observer_script }}.j2"
+        dest: "{{observer_script_dir}}/{{ observer_script }}"
+        owner: "{{ oracle_user }}"
+        group: "{{ oracle_group }}"
+        mode: "ug=rwx,o="
+
+    - name: Start the observer
+      shell: "{{observer_script_dir}}/{{ observer_script }}"
+
+    - name: Wait for a few seconds
```
mfielding: Rather than a static wait, an ansible-style polling look with until/retries/delay might be more flexible, and a bit faster too.

docs/disaster-recovery-configurations.md
```diff
@@ -142,6 +181,18 @@ Customize the Oracle databases and Data Guard configurations provisioned using t
...
+
+1. Supported for Oracle Database versions 19c and above.
+
+2. The playbook creates a user with `SYSDG` privilege, and its password is stored in an auto-login  wallet on the FSFO host.
+
+3. FSFO provisioning re-run will change the `SYSDG` user password and will update the wallet password as well.
+
+4. The wallet password is stored in the `oracle-toolkit.tmp` file. Check it out and remove it after the deployment.
+
+5. The FSFO observer startup has be handled manually
```
mfielding: There's a typo here, but if we automate the startup via systemd, it becomes a moot point.

config-observer.yml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+- name: Configure Data Guard observer
+  hosts: dbasm
+  gather_facts: false
+  roles:
+    - role: dg-observer
```
mfielding: nit: missing trailing newline

docs/disaster-recovery-configurations.md
pythianakhmadeev: I'm adding `--ora-edition CLIENT` option to allow installation of an Oracle Client, if it's required.

docs/disaster-recovery-configurations.md
pythianakhmadeev: I've removed HugePages configuration for a CLIENT installation. Disk prep will stay, as having a separate mount for Oracle home is a standard thing.

docs/disaster-recovery-configurations.md
pythianakhmadeev: This is not a straight forward thing to do, as far as I can see.

roles/dg-observer/tasks/main.yml
pythianakhmadeev: Yes we would. That was initially done to allow re-run of the Observer provisioning.

mfielding (commented):
Here are some suggestions to optimize the client-mode installation (for DG Observer) and avoid potential issues.

### General Suggestions (for files not in the diff):

**1. Optimize `patch` role for Client:**
To support patching the client home without failing (due to missing database instance), we should modify `roles/patch/tasks/main.yml` and `roles/patch/tasks/opatch_apply.yml`:
*   Skip database role checks (`sqlplus / as sysdba`) when `ora_edition == 'CLIENT'`.
*   Skip `srvctl stop/start home` tasks for client homes.
*   Skip SQL patching (`datapatch`) in `opatch_apply.yml` for client.

**2. Avoid creating `grid` user for Client:**
Currently, the `grid` user is still created because `oracle_users` in `roles/common/defaults/main.yml` is not modified. If we want to fully optimize, we should make `grid` user creation conditional. Note that if we do this, we also need to adjust `oracle_dirs` in `roles/common/defaults/main.yml` because `oracle_inventory` is currently configured to be owned by `grid_user` by default, which would fail if the user doesn't exist.

roles/ora-host/tasks/main.yml
```diff
@@ -198,6 +198,7 @@
     file: hugepages.yml
     apply:
       tags: hugepages
+  when: not (ora_edition | lower == 'client')
```
mfielding: Good addition. We should similarly skip other DB-specific tasks in this file (like sysctl tuning, `/dev/shm` mount, and limits for `grid` user) when `ora_edition` is `CLIENT`.

roles/rdbms-setup/templates/db_install.rsp.sh.j2
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+RSP_FILE={% if ora_edition | lower == "client" %}"clientsetup.rsp"{% else %}"db_install.rsp"{% endif %}
```
mfielding: Since we are using `clientsetup.rsp` here, we should also ensure we set `oracle.install.client.installType=Administrator` later in this file (currently it tries to set `oracle.install.db.InstallEdition` which is invalid for client). Also, we should skip other DB-specific `sed` commands (like OS group mappings) when `ora_edition == 'CLIENT'`.

roles/common/defaults/main/rdbms_software.yml
```diff
@@ -22,6 +22,10 @@ rdbms_software:
     files:
       - { name: "V1054592-01.zip", sha256sum: "2a5d2583df076209e624dff750829f9a562473c27c8f926d546cf9ad2a1a2efd", md5sum: "kfS0nf1YbfK8kEq2Gu3v9w==",
           alt_name: "LINUX.X64_2326100_db_home.zip", alt_sha256sum: "2a5d2583df076209e624dff750829f9a562473c27c8f926d546cf9ad2a1a2efd", alt_md5sum: "kfS0nf1YbfK8kEq2Gu3v9w==" }
+  - name: 26ai_base_client
+    version: 23.26.1.0.0
+    edition: CLIENT
+    files: { name: "LINUX.X64_2326100_client_home.zip", sha256sum: "55B8D08F9AB77C0BCEA1CD51DAE6F4EE306112E02B44063468CC2EF3D8AD4BCE", md5sum: "qEkZ99edVeO2lr7+LAzwXg==" }
```
mfielding: Can you add the eDelivery name of the client too?

Also, do you mind updating the check_sw playbook to cover client mode?

install-oracle.sh
```diff
@@ -241,11 +247,12 @@ if [ "$HELP_ONLY" = true ]; then
   echo "  --install-sw                 Run only the software installation playbook and exit."
   echo "  --no-patch                   Set Oracle release to 'base' (skip patching)."
   echo "  --config-db                  Run only the database configuration playbook and exit."
+  echo "  --config-observer            Run the Data Guard Observer installation and configuration playbooks."
   echo "  --skip-database-config       Skip database configuration."
   echo "  --install-workload-agent     Install the workload agent."
   echo "  --ora-version <version>      Oracle version (e.g., 19, 23, 26)."
   echo "  --ora-release <release>      Oracle release (e.g., 19.0.0.0.0)."
-  echo "  --ora-edition <edition>      Oracle edition (EE, SE, SE2, FREE)."
+  echo "  --ora-edition <edition>      Oracle edition (EE, SE, SE2, FREE, CLIENT)."
```
mfielding, line 259: Note that running `install-oracle.sh` with `--ora-edition CLIENT`, but without `--config-observer` would fail, as it would attempt to create a database.

To fully support CLIENT as a DB edition, we probably want some logic to avoid calling ${PB_CONFIG_DB} or ${PB_COMPATIBLE}

mfielding: Or, alternately, maybe we don't need `--ora-edition CLIENT` at all, but rather checking for some flag that `--config-observer` would set.
I guess I'm trying to either have full and tested client (non-observer) installs, or leave the client installs as a special case for observers only.

docs/disaster-recovery-configurations.md
```diff
@@ -47,6 +49,25 @@ export STANDBY_IP_ADDR=10.0.10.102
...
+export PRIMARY_IP_ADDRESS=10.0.10.101
+
+./install-oracle.sh \
+  --config-observer \
+  --cluster-type DG \
+  --instance-ip-addr ${STANDBY_IP_ADDR} \
+  --primary-ip-addr ${PRIMARY_IP_ADDRESS} \
+  --instance-hostname standby-server-19c \
+  --ora-version 19 --ora-edition CLIENT \
+  --ora-disk-mgmgt FS \
```
mfielding: Nit: typo in --ora-disk-mgmt

docs/user-guide.md
```diff
@@ -2426,6 +2426,14 @@ server.</td>
 <tr>
 <td></td>
 <td><p><pre>
+--config-observer
+</pre></p></td>
+<td></td>
+<td>Run the "config-observer.yml" playbook only.</td>
```
mfielding: With the latest changes, I'd say:

Run the Data Guard Observer installation and configuration playbooks.

install-oracle.sh
```diff
@@ -197,6 +199,10 @@ if [[ -z "${YAML_VARS[cluster_config_json]}" && -n "${YAML_VARS[cluster_config]}
   YAML_VARS["cluster_config_json"]="${JSON_CONTENT}"
   unset YAML_VARS["cluster_config"]
 fi
```
mfielding, line 201: Consider defaulting `ora_edition` to `CLIENT` when the observer is configured:

```
if [ "$CONFIG_OBSERVER" = true ] && [[ -z "${YAML_VARS[ora_edition]}" ]]; then
  YAML_VARS["ora_edition"]="CLIENT"
fi
```

mfielding (approved):

PR #450, opened 2026-09-10
fix(terraform): normalize numeric inputs and prevent phantom memory flags

mfielding:
### Summary
Fixes string comparison antipattern (`var.X != ""`) on numeric Terraform variables in `terraform/main.tf` and `terraform/variables.tf`.

### Problem Statement
1. **Phantom CLI Flags (`0 != ""` Coercion):**  
   In `terraform/variables.tf`, `ora_pga_target_mb` and `ora_sga_target_mb` default to `0` to signal that the toolkit should automatically compute memory (45% of host RAM for SGA, HugePages allocation, and baseline PGA). However, `terraform/main.tf` checks:
   ```hcl
   var.ora_pga_target_mb != "" ? "--ora-pga-target-mb ${var.ora_pga_target_mb}" : "",
   var.ora_sga_target_mb != "" ? "--ora-sga-target-mb ${var.ora_sga_target_mb}" : "",
   ```
   In HCL, comparing a number to `""` performs implicit string coercion: `"0" != ""` evaluates to `true`. This causes `--ora-pga-target-mb 0` and `--ora-sga-target-mb 0` to **always** be passed on standard deployments when unset by the caller.
2. **Runtime Errors on `null`:**  
   When callers pass `null` into module arguments, `null != ""` evaluates to `true`, and string template interpolation (`"${var.ora_pga_target_mb}"`) aborts with:
   `Invalid interpolation: null value cannot be formatted as a string`.
3. **Redo Log Count Typing & Validation:**  
   `ora_redo_log_count` was previously typed as `string` with a regex check (`^[0-9]+$`) to work around the string idiom. Passing `null` fails validation because `null == ""` is `false` and regex fails `can()`.
4. **Port Number Validation:**  
   `ora_listener_port` and `tls_listener_port` validation rules (`var.port >= 1`) crash if a caller passes `null`.

### Solution
1. **Replace String Checks with `try(var.X > 0, false)`:**  
   In `local.common_flags`, replace `var.num != ""` with `try(var.num > 0, false)` for `ora_pga_target_mb`, `ora_sga_target_mb`, `ora_listener_port`, `tls_listener_port`, and `ora_redo_log_count`.
2. **Null-Safe Validations:**  
   Update validation conditions across all numeric inputs in `variables.tf` to explicitly permit `null`.
3. **Flexible `ora_redo_log_count`:**  
   Change `ora_redo_log_count` type to `any` with validation allowing numbers, strings (e.g. `"2"` in existing tfvars templates), empty string, or `null`.
4. **Firewall Port Casting:**  
   Add `tostring()` to firewall rule port mapping to ensure strict schema compliance.

### Test Results & Matrix
A verification test suite was executed across all 7 caller scenarios (defaults, explicit tuning, explicit zero, nulls, custom ports, string numbers, and TLS). 

The full test matrix output is published in this GitHub Gist:
https://gist.github.com/mfielding/e7c3a4e9db5bd7594ff6d00309fe3a22


terraform/main.tf
```diff
@@ -551,19 +551,19 @@ locals {
     var.ntp_pref != "" ? "--ntp-pref ${var.ntp_pref}" : "",
     var.ora_release != "" ? "--ora-release ${var.ora_release}" : "",
     var.ora_edition != "" ? "--ora-edition ${var.ora_edition}" : "",
-    var.ora_listener_port != "" ? "--ora-listener-port ${var.ora_listener_port}" : "",
+    try(var.ora_listener_port > 0, false) ? "--ora-listener-port ${var.ora_listener_port}" : "",
```
AlexBasinov: should we add tonumber() here as well?

terraform/main.tf
```diff
@@ -551,19 +551,19 @@ locals {
...
-    var.ora_redo_log_count != "" ? "--ora-redo-log-count ${var.ora_redo_log_count}" : "",
+    try(tonumber(var.ora_redo_log_count) > 0, false) ? "--ora-redo-log-count ${var.ora_redo_log_count}" : "",
     var.ora_redo_log_location != "" ? "--ora-redo-log-location '${var.ora_redo_log_location}'" : "",
     var.db_password_secret != "" ? "--db-password-secret ${var.db_password_secret}" : "",
     var.oracle_metrics_secret != "" ? "--oracle-metrics-secret ${var.oracle_metrics_secret}" : "",
     var.install_workload_agent ? "--install-workload-agent" : "",
     var.skip_database_config ? "--skip-database-config" : "",
-    var.ora_pga_target_mb != "" ? "--ora-pga-target-mb ${var.ora_pga_target_mb}" : "",
-    var.ora_sga_target_mb != "" ? "--ora-sga-target-mb ${var.ora_sga_target_mb}" : "",
+    try(var.ora_pga_target_mb > 0, false) ? "--ora-pga-target-mb ${var.ora_pga_target_mb}" : "",
```
AlexBasinov: and here as well

terraform/main.tf
mfielding: Added.

terraform/main.tf
mfielding: I actually added it to all the numerical checks.

AlexBasinov (approved):

